PR #7, opened 2015-08-28
Add a proxmox-agent

tuxis-ie:
The LibreNMS-side will follow..


f0o:
Will the LibreNMS backend side include docs on how to config this? (if it requires any config)

tuxis-ie:
Yes, this end needs no configuration other than `mk_enplug proxmox`

PR #25, opened 2016-05-05
Improved hddtemp agent module

exarv:
I had some issues with the netcat / daemon implementation of the module.
netcat was stalling or sometimes netcat did not return the full output of hddtemp.
Running hddtemp directly without running it as a daemon is much more stable for me.

This new version also does not give any stdout output when hddtemp is not installed or when no disks can be found.
Running the script manually on a server does give stderr output for easy debugging.


f0o:
Well done :+1:

PR #29, opened 2016-06-30
Python3 fixes for powerdns agent.  Compatible with python2.

murrant:


paulgear:
Looks good to me, but I don't have any PowerDNS to test against.

PR #30, opened 2016-07-09
Updates to MySQL Agent, to work with PHP7 (mysql -> mysqli)

arrmo:
To this issue,
https://github.com/librenms/librenms/issues/3751#issuecomment-231546464

Yell if you seen anything wrong. Thanks!

### DIFF
--- a/agent-local/mysql
+++ b/agent-local/mysql
@@ -49,7 +49,7 @@ $mysql_user = '';
 $mysql_pass = '';
 $mysql_host = 'localhost';
 $mysql_port = 3306;
-$mysql_ssl  = FALSE;   # Whether to use SSL to connect to MySQL.
+$mysql_ssl  = FALSE;    # Whether to use SSL to connect to MySQL.
 
 $heartbeat   = '';      # db.tbl in case you use mk-heartbeat from Maatkit.
 $cache_dir  = '/tmp';   # If set, this uses caching to avoid multiple calls.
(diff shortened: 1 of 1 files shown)


agent-local/mysql
```diff
@@ -270,18 +270,18 @@ function ss_get_mysql_stats( $options ) {
...
-      $conn = mysql_connect($host_str, $user, $pass, true, MYSQL_CLIENT_SSL);
+      $conn = mysqli_connect($host_str, $user, $pass, true, MYSQL_CLIENT_SSL);
    }
    else {
-      $conn = mysql_connect($host_str, $user, $pass);
+      $conn = mysqli_connect($host_str, $user, $pass);
    }
    if (!$conn ) {
-      die("MySQL: " . mysql_error());
+      die("MySQL: " . $conn->connect_error);
```
paulgear: How are we going to get $conn->connect_error if $conn is null?

arrmo: Good catch! My mistake. Checking out this, http://php.net/manual/en/mysqli.error.php, and the code there,
if ($mysqli->connect_errno) {
    printf("Connect failed: %s\n", $mysqli->connect_error);
    exit();
}

Seems the code above should be as follows, agreed?
if ($conn->connect_errno) {

I admit - I had it running, but missed checking with MySQL down. Again, thanks!

laf: yeah that looks good.

laf:
@arrmo thanks for doing this, @murrant also submitted a fix and as that didn't need any further changes we've merged that one.

arrmo:
NP - works for me. Thanks!

PR #40, opened 2016-08-03
Add Debian and make update call more robust

florianbeer:
- Debian based systems need to update the index before being able to report upgradable packages.
- Debian old-stable doesn't have `apt` yet and Ubuntu 14.04 emits the following warning when using `apt` in a script:
`WARNING: /usr/bin/apt does not have a stable CLI interface yet. Use with caution in scripts.`

By using `apt-get`, issuing a `update` call first and then counting the result of `grep 'Inst'`, this script now works on Debian 7, Debian 8, Ubuntu 14.04 and Ubuntu 16.04.

### DIFF
--- a/snmp/os-updates.sh
+++ b/snmp/os-updates.sh
@@ -12,8 +12,9 @@ BIN_ZYPPER='/usr/bin/zypper'
 CMD_ZYPPER='lu'
 BIN_YUM='/usr/bin/yum'
 CMD_YUM='check-update'
-BIN_APT='/usr/bin/apt'
-CMD_APT='list --upgradable'
+BIN_APT='/usr/bin/apt-get'
+CMD_APT='-s upgrade'
+CMD_UPDATE='-qq update'
 
 #general check for os based on /etc/os-release
 if [ -f /etc/os-release ]; then
@@ -32,10 +33,11 @@ if [ -f /etc/os-release ]; then
 		else
 			echo "0";
 		fi
-	elif [ $OS == "ubuntu" ]; then
-		UPDATES=`$BIN_APT $CMD_APT | $BIN_WC $CMD_WC`
+	elif [ $OS == "ubuntu" ] || [ $OS == "debian" ]; then
+		`$BIN_APT $CMD_UPDATE`
+		UPDATES=`$BIN_APT $CMD_APT | grep 'Inst' | $BIN_WC $CMD_WC`
 		if [ $UPDATES -gt 1 ]; then
-			echo $(($UPDATES-1));
+			echo $UPDATES;
 		else
 			echo "0";
 		fi
--- a/snmp/os-updates.sh
+++ b/snmp/os-updates.sh
@@ -14,7 +14,6 @@ BIN_YUM='/usr/bin/yum'
 CMD_YUM='check-update'
 BIN_APT='/usr/bin/apt-get'
 CMD_APT='-s upgrade'
-CMD_UPDATE='-qq update'
 
 #general check for os based on /etc/os-release
 if [ -f /etc/os-release ]; then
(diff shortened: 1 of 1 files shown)


snmp/os-updates.sh
```diff
@@ -32,10 +33,11 @@ if [ -f /etc/os-release ]; then
 		else
 			echo "0";
 		fi
-	elif [ $OS == "ubuntu" ]; then
-		UPDATES=`$BIN_APT $CMD_APT | $BIN_WC $CMD_WC`
+	elif [ $OS == "ubuntu" ] || [ $OS == "debian" ]; then
+		`$BIN_APT $CMD_UPDATE`
```
paulgear: Why the backquotes on this?  If you don't want the output, just >/dev/null - it's clearer.

florianbeer: You're absolutely right. Never code while tired!

florianbeer:
Actually we have a bit of a problem here, `apt-get -qq update` needs to lock the directory `/var/lib/dpkg/` which it only can do as root. Now if you're a sane sysadmin, you're not running snmpd as root, so this won't work.

What we can do, is rely on an external mechanism to update the index. Debian recommends creating `/etc/apt/apt.conf.d/10periodic` and setting `APT::Periodic::Update-Package-Lists "1";`. If you have `apticron`, `cron-apt` or `apt-listchanges` installed and configured, chances are that packages are already updated periodically. I found the setting in the following files in different Debian and Ubuntu versions:

- `/etc/apt/apt.conf.d/20auto-upgrades`
- `/etc/apt/apt.conf.d/10periodic`

Should we attempt to activate this settings automatically in the script, or would it be better to just make a note in the docs? Your opinions please.

tuxis-ie:
Wouldn't it make more sense to use the unix-agent for this? That already runs as root and also manages `dpkg`.
Please do not change any settings for a script.

florianbeer:
I don't quite see the need for installing another agent program and opening up more ports on the host, if snmp already offers agentx, extend etc. I like to keep things simple and poll all data through snmp.

tuxis-ie:
If you are installing this package on a debian-box, you already have the agent, the open port, etc. :)

florianbeer:
But not on remote machines.  
What snmp extend lets us do is, snmpget a specific OID under NET-SNMP-EXTEND-MIB which is defined by the `extend` line in snmpd.conf. Then snmp executes a local script and returns the value. So all configuration that is necessary after deploying the script on a remote server is adding one line to snmpd.conf and reloading snmpd. No additional binaries, no additional ports etc.

Here's a link with some more info: http://sysadvent.blogspot.co.at/2008/12/day-4-extending-net-snmps-snmpd.html

Btw. this already works perfectly for the inlcuded apache, mysql and nginx "Applications", just deploy the script from this repo (e.g. https://github.com/librenms/librenms-agent/blob/master/agent-local/apache) on the remote machine, add the extend line (e.g. `extend apache /usr/local/bin/apache`) and reload the snmp daemon. LibreNMS will happily graph all arriving data, because it already checks via snmpget (https://github.com/librenms/librenms/blob/master/includes/polling/applications/apache.inc.php#L7:L11)

To me this is far more desirable than installing the Agent (http://docs.librenms.org/Extensions/Applications/#agent-setup)

laf:
I'd say not to make local changes via a script.

net-snmp runs as root by default doesn't it - who realistically changes that? I'd inclined to say leave it as is but if you have to go down the road of having people make further changes then you'll need to document it well - we've learned that the harder it is the less likely people can use it out of the box.

florianbeer:
I will remove the update code and add documentation stating, that Ubuntu and Debian depend on automatic package index updates.

PR #42, opened 2016-08-03
app-dhcp-stats snmp extend

crcro:
local snmp extend script required for dhcp stats app


paulgear:
Not important, but that script could be made a lot prettier by using a loop instead of repeating the same code... :-)

PR #48, opened 2016-08-14
raspberry sensors

crcro:
added snmp extend to get raspberry sensors

### DIFF
--- a/snmp/raspberry.sh
+++ b/snmp/raspberry.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+#######################################
+# please read DOCS to succesfully get #
+# raspberry sensors into your host    #
+#######################################
+picmd='/usr/bin/vcgencmd'
+pised='/bin/sed'
+getTemp='measure_temp'
+getVoltsCore='measure_volts core'
+getVoltsRamC='measure_volts sdram_c'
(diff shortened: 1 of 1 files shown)


snmp/raspberry.sh
```diff
@@ -0,0 +1,34 @@
...
+getFreqArm='measure_clock arm'
+getFreqCore='measure_clock core'
+getStatusH264='codec_enabled H264'
+getStatusMPG2='codec_enabled MPG2'
+getStatusWVC1='codec_enabled WVC1'
+getStatusMPG4='codec_enabled MPG4'
+getStatusMJPG='codec_enabled MJPG'
+getStatusWMV9='codec_enabled WMV9'
+
+sudo $picmd $getTemp | $pised 's|[^0-9.]||g'
```
murrant: Is sudo required?

crcro: unfortunately yes, because vcgencmd won't return any data if not run by root

murrant: If any extra configuration is needed to allow sudo to run, make sure you add it to the docs in you PR on the main repo.

crcro: i did that already on main repo with PR #4057

PR #51, opened 2016-08-21
rewrite dhcp-stats with loop

crcro:
@paulgear recomandation in PR #42


paulgear:
Much prettier. :-) :+1:

PR #56, opened 2016-09-08
Update hddtemp to include hddtemp -w option

vectr0n:
hddtemp gives inconsistent values in it's current state, after some debugging I was able to resolve the issue by passing -w to the hddtemp command, this will wake-up the drive if it is in a sleep state to gather information.


laf:
Thanks @vectr0n 

image

PR #58, opened 2016-09-10
snmp-extend: ups-apcups

crcro:
shell script to export linux apc ups app status

### DIFF
--- a/snmp/ups-apcups.sh
+++ b/snmp/ups-apcups.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+################################################################
+# copy this script to somewhere like /opt and make chmod +x it #
+# edit your snmpd.conf and include                             #
+# extend ups-apcups /opt/ups-apcups.sh                         #
+# restart snmpd and activate the app for desired host          #
+# please make sure you have the path/binaries below            #
+################################################################
+# Binaries and paths required                                  #
+################################################################ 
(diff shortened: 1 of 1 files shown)


snmp/ups-apcups.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+################################################################
+# copy this script to somewhere like /opt and make chmod +x it #
+# edit your snmpd.conf and include                             #
+# extend ups-apcups /opt/ups-apcups.sh                         #
```
laf: We need to standardise the install directory - people can change it but we should tell those that don't know to / want to. I think we have suggestions in the docs of where to store these, use one of those.

crcro: The suggestion from docs is for agent setup, maybe we should put this straight now, and I'll change the docs and all other. I suggest something like /opt/librenms/apps or /etc/snmpd/extend

laf: /etc/snmp/ seems to be used regularly so I'd go with that - @murrant ?

murrant: That is why I used that in my documentation entries, so I would prefer /etc/snmp

laf:
Just needs the header changing of where to install this to match docs then good to go.

crcro:
@laf .. done.

PR #61, opened 2016-09-12
app: ntp-server update 1

crcro:
added a fix for older versions that does not support ntpq -c iostats


murrant:
Why use the full path for these commands?  Does that make it less likely to work without editing?

PR #62, opened 2016-09-14
snmp-extend: ntp-server update 2

### LINKED ISSUE #43: ntpd stats non functional
poller -d run:
http://pastebin.com/kEgbJGB2

crcro:
- easier to maintain code
- new snmp location as per agreed
- fix #43

### DIFF
--- a/snmp/ntp-server.sh
+++ b/snmp/ntp-server.sh
@@ -23,8 +23,7 @@ TMP_FILE1='/tmp/output-ntpserver'
 # Don't change anything unless you know what are you doing     #
 ################################################################
 VER=`$BIN_NTPD --version`
-
-if [[ "$VER" =~ '4.2.6p5' ]]
+if [[ $VER =~ .*4.2.6p5.* ]]
 then
   USECMD=`echo $BIN_NTPDC -c iostats`
 else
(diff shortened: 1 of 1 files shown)


crcro:
@laf ... this can be merged also

laf:
I'm not an ntp user really.

@librenms/reviewers anyone else?

Gorian:
I can test this when I get a chance. One thing I wasn't sure - is this for ntpd, running as a server to sync the time, or for a server SERVING the time to ntp clients?

rbax82:
This is serving the time to clients

paulgear:
FWIW, ntpd syncing its own time and serving the time to NTP clients are the same program; the only difference is what firewall rules you apply, and what restrictions are in /etc/ntp.conf.

rbax82:
ntpd 4.2.6p5
***Command `iostats' unknown
2
-0.515
0.119
2.770
2.909
0.748

rbax82:
Also grep and ntpq are in different location under Ubuntu
BIN_GREP='/bin/grep'
and
BIN_NTPQ='/usr/bin/ntpq'

paulgear:
I don't use the agent for NTP monitoring (I use collectd + my ntpmon script instead), but this change is -1 from me.  I don't think we should be using this bash coding style; it makes things less portable, and encourages poor use of shell features.  I'm happy to help with making it use a more familiar & portable form of shell coding.

tsumaru720:
@paulgear as it happens I'd submitted my own PR to replace these scripts (before I knew this one existed) - see #110 - which would you prefer?

paulgear:
@tsumaru720 Definitely an improvement, although why you would want to use `which` to set variables pointing to the scripts rather than just let `$PATH` select the right ones doesn't make much sense to me.

tsumaru720:
I figured it'd save hardcoding them like the old one does (as the hardcoded values didnt work on Raspbian) - Do you think they're even neded @paulgear ?

Aka get rid of the binary variables all together?

addp009:
Any outstanding issue to address here? I started using this script in master and was getting wrong results as described by @tsumaru720 in https://github.com/librenms/librenms-agent/issues/109. 

Tested with the fixes in this PR and works correctly.

Only suggestion is that not hardcoding the path to the binaries.  I have a separate PR for that: https://github.com/librenms/librenms-agent/pull/150

SourceDoctor:
2 years nothing happens
also completed outdated
running version returns also JSON

Issue #64, opened 2016-09-16
os-updates broken on OpenSuSE Leap


Tatermen:
snmp/os-updates.sh attempts to determine the distribution by checking for the existence of the various package managers, eg. zypper for OpenSuSE, yum for Redhat/CentOS and apt-get for Debian.

OpenSuSE Leap installs an apt-get wrapper that attempts runs the equivalent commands in zypper, but fails to do so correctly in this case.

On OpenSuSE 13.2, os-updates.sh runs the following command and gets a list of updates:

```
# zypper lu
S | Repository                | Name            | Current Version | Available Version | Arch
--+---------------------------+-----------------+-----------------+-------------------+-------
v | openSUSE-Leap-42.1-Update | apache2         | 2.4.16-12.1     | 2.4.16-15.1       | x86_64
v | openSUSE-Leap-42.1-Update | apache2-devel   | 2.4.16-12.1     | 2.4.16-15.1       | x86_64
v | openSUSE-Leap-42.1-Update | apache2-prefork | 2.4.16-12.1     | 2.4.16-15.1       | x86_64
v | openSUSE-Leap-42.1-Update | apache2-utils   | 2.4.16-12.1     | 2.4.16-15.1       | x86_64
v | openSUSE-Leap-42.1-Update | libtiff5        | 4.0.6-3.1       | 4.0.6-6.1         | x86_64
v | openSUSE-Leap-42.1-Update | perl            | 5.18.2-3.5      | 5.18.2-5.1        | x86_64
v | openSUSE-Leap-42.1-Update | perl-base       | 5.18.2-3.5      | 5.18.2-5.1        | x86_64
```

On OpenSuSE Leap, it instead does ```apt-get -s upgrade```, which the crappy script interprets as ```zypper --dry-run update```:

```
# apt-get -s upgrade
Loading repository data...
Reading installed packages...

The following 7 packages are going to be upgraded:
  apache2 apache2-devel apache2-prefork apache2-utils libtiff5 perl perl-base

7 packages to upgrade.
Overall download size: 9.2 MiB. Already cached: 0 B. After the operation, 20.1 KiB will be freed.
Continue? [y/n/? shows all options] (y):
```
and waits for input from a non-existent user.

This could be fixed by re-writing os-updates.sh to instead read /etc/os-release to determine the distribution and run the appropriate package manager. I believe this file should exist on any modern systemd based distribution - but may cause problems with older distros.

crcro:
[...]

PR #65, opened 2016-09-16
snmp-extend: os-updates

crcro:
- reverted back to os-release checks (fix issue: #64)
- added arch pacman support


crcro:
@laf ... this can be merged also

PR #69, opened 2016-10-08
Agent: script to collect data from GPSD

karlshea:


agent-local/gpsd
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/php
```
laf: Replace this with: `#!/usr/bin/env php`

PR #71, opened 2016-11-17
Add Debian support

matunixe:
Before that only Ubuntu was supported, now Debian users can use this nice script too!


matunixe:
Thanks! 👍

PR #74, opened 2016-12-08
added fedora support

bungojungo:
Added support for detecting Fedora. As Fedora has an /etc/redhat-release, I added the check for /etc/fedora-release before /etc/redhat-release.


bungojungo:
Closing, will-reopen after fix.

PR #77, opened 2016-12-20
asterisk check script

rbax82:
Adding support to the agent to support asterisk


laf:
Does this require any files on the librenms/librenms repo for graphing?

rbax82:
Yes it does and it appears I don't have the required knowledge to write them. :/

PR #78, opened 2017-01-08
Tchmnkyz patch 1

tchmnkyz:
I had problems with the scripts working on some debian based systems. To fix it i used the following. Thought I would share it so others might not have the same problem.

### DIFF
--- a/snmp/ntp-client.sh
+++ b/snmp/ntp-client.sh
@@ -8,10 +8,10 @@
 ################################################################
 # Binaries and paths required                                  #
 ################################################################ 
-BIN_NTPQ='/usr/sbin/ntpq'
-BIN_GREP='/usr/bin/grep'
-BIN_TR='/usr/bin/tr'
-BIN_CUT='/usr/bin/cut'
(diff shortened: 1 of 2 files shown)


crcro:
@tchmnkyz ... those value should be changed by each user individually based on their distro layout, changing them as you did it would render them useless on other os distros.

tchmnkyz:
@crcro, Actually i checked multiple distro's before i submitted this. Freebsd, Debian, CentOS, Redhat all had the same path to which. This means that it should work on most distributions without too many issues.

murrant:
What is the point of using which here?  Just put the commands without the the path and it does exactly the same thing.

Also, why do we use full path in the first place?

PipoCanaja:
Should we continue in this PR or just remove full path here and trust SNMPd PATH definition to find the binaries ?

SourceDoctor:
Binarypathes can be overwritten if needed

`...
CONFIG=$0".conf"
if [ -f $CONFIG ]; then
    . $CONFIG
fi
....
`
so no need for PR

PR #89, opened 2017-02-24
Dynamic library naming in shell scripts

mrcnpdlk:
To each bash file I added code like below.
This check if binary exists and return variable BIN_* with path to file.
In my opinion is better than siple using "$(command -v some_binary_name)"

```bash
exists_command()
{
  command -v "$1" >/dev/null 2>&1
}

require_commands=("cat" "grep" "tr" "sed" "sort" "wc") ;

for i in "${require_commands[@]}" ;
do
    if exists_command $i; then
        eval "BIN_${i^^}"="$(command -v $i)";
    else
        echo "Your system does not have [$i]";
        exit
    fi

done ;
```


SourceDoctor:
Not the best way for my intention.
Optional configfile which overrides binaries with systemspecific pathes will be the easier and faster way instead of running in each script for each binary a which command.
For example take a look here: [dhcp-stats.sh](https://github.com/librenms/librenms-agent/blob/master/snmp/dhcp-status.sh)

PR #94, opened 2017-02-27
add Nvidia SNMP extend poller

VVelox:
Adds support for tracking Nvida GPU stats.


VVelox:
Updated this so it will now work properly if more than 4 GPUs are installed on a sytem.

PR #97, opened 2017-03-06
make ntp-client.sh dd-wrt(with ipkg) friendly and remove bash requirement

VVelox:
Yes, I've signed the author stuff etc... For some reason this form is not present here...

Wooho!

For DD-WRT, install ntp via ipkg as ntpq does not come in the base.

root@wap0:/jffs/snmpd# ./ntp-client
50.060777
-8.028
29.335078
8.745
0.012
root@wap0:/jffs/snmpd#

### DIFF
--- a/snmp/ntp-client.sh
+++ b/snmp/ntp-client.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 ################################################################
 # copy this script to somewhere like /opt and make chmod +x it #
 # edit your snmpd.conf and include                             #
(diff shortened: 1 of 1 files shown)


VVelox:
Rewrote with out the bash stuff as for some reason it was choking on that with the version of bash available via ipkg. Tested on FreeBSD and this works there as well.

rockyluke:
On Debian and Ubuntu, **/bin/sh** is ash which is very limited. Maybe we need to find a better way to be full compatible for BSD and Linux ?

VVelox:
No advanced shell stuff is required and this was tested on ash(DD-WRT). Also this is not a BSD v. Linux issue. This is specifically that it does not work with DD-WRT and the version of bash available via ipkg.

rockyluke:
That's not what I meant...

That's OK for me :-)

VVelox:
Okay. I am confused then. =^.^=

PR #103, opened 2017-03-09
add BIND named SNMP extend

VVelox:
for a soon to be update to the BIND named stats poller...


VVelox:
actually closing this for now... even with the last stat dump, blowing way to much info over the wire

PR #105, opened 2017-03-11
Postfix app bug fixes

florianbeer:
- add missing DS
- fix some totals

### DIFF
--- a/snmp/postfixdetailed
+++ b/snmp/postfixdetailed
@@ -500,10 +500,10 @@ my $data=$received."\n".
 	$rardnf."\n".
 	$rarnfqa."\n".
 	$iuscp."\n".
-	$msefl."\n".
 	$sce."\n".
 	$scp."\n".
 	$urr."\n";
+	$msefl."\n".
 
 print $data;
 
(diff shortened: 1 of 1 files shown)


laf:
That seems to replace some ds' - what's the issue this pr is trying to fix?

florianbeer:
There where multiple issues here:

Not all data sources produced meaningful data because they referred to variables that where missing at some other points in the script. There was one line check that had the variable names from the previous check (likely copy & paste error). This resulted in values being printed out in the wrong position for the expected snmp check or not being filled with data at all.

Secondly I added two more line checks that allowed for multiple occurrences (like RBL), based on empirical evidence from my log files (doesn't change anything if there is only one match). This messed up the old vs. new value calculation, resulting in data sources that always had the same value at every data point in time.

In general it is maybe wise to think about refactoring this whole app, since it uses GAUGE for its rrd's. pflogsumm already reports increasing values (until it rolls over after maillog logrotate), this would indicate the usage of the COUNTER data type. When using COUNTER this snmp script could skip writing a cache file and all internal calculations and just postprocess pflogsumm output and print it to stdout.

PR #106, opened 2017-03-12
actually track $msefl now and also properly track $iuscp

VVelox:
Found this in https://github.com/librenms/librenms/pull/6173 .

### DIFF
--- a/snmp/postfixdetailed
+++ b/snmp/postfixdetailed
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/local/bin/perl
 
 # add this to your snmpd.conf file as below
 # extend postfixdetailed /etc/snmp/postfixdetailed
@@ -7,7 +7,7 @@
 my $cache='/var/cache/postfixdetailed';
 
 # the location of pflogsumm
-my $pflogsumm='/usr/bin/env pflogsumm';
+my $pflogsumm='/usr/local/bin/pflogsumm';
 
 #totals
 #    847   received                 = received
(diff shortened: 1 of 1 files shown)


VVelox:
Much thanks to @florianbeer for the suggestions. :)

snmp/postfixdetailed
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/local/bin/perl
```
laf: What's up with using env here?

snmp/postfixdetailed
```diff
@@ -7,7 +7,7 @@
 my $cache='/var/cache/postfixdetailed';
 
 # the location of pflogsumm
-my $pflogsumm='/usr/bin/env pflogsumm';
+my $pflogsumm='/usr/local/bin/pflogsumm';
```
laf: Same again

VVelox:
@florianbeer Whoops. Yeah. Copied and paste of it from the system I was testing on did that. Fixed now.

@laf Fixed.

VVelox:
Also, no, this does not break any existing systems as this appends the variable to the end. As to the other changes, those just make sure stuff is properly counted.

laf:
Why not just have the small change done in florians PR?

VVelox:
Could not make the changes, it would of broken stuff because of the positioning of $msefl during the print, and I also found another small bug as well.

laf:
But it would be better to just collaborate with him on his PR :)

PR #107, opened 2017-03-15
Added try-except checks for global values.

BlackDex:
Fixed an error which prevented output.
It seems some ceph version probably use different values or something. This is a quick fix to have the script output the correct values.

### DIFF
--- a/agent-local/ceph
+++ b/agent-local/ceph
@@ -21,7 +21,20 @@ def cephdf():
     cephdf    = check_output(["/usr/bin/ceph", "-f", "json", "df"]).replace('-inf', '0')
 
     s = json.loads(cephdf)
-    print("c:%i:%i:%i" % (s['stats']['total_bytes'], s['stats']['total_used_bytes'], s['stats']['total_avail_bytes']))
+    try:
+        ts = s['stats']['total_bytes']
+    except:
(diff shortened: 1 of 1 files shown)


agent-local/ceph
```diff
@@ -21,7 +21,20 @@ def cephdf():
...
-    print("c:%i:%i:%i" % (s['stats']['total_bytes'], s['stats']['total_used_bytes'], s['stats']['total_avail_bytes']))
+    try:
+        ts = s['stats']['total_bytes']
+    except:
+        ts = s['stats']['total_space']
+    try:
+        tu = s['stats']['total_used_bytes']
+    except:
+        tu = s['stats']['total_used']
+    try:
```
murrant: Why doesn't this script just output the json and handle everything in LibreNMS...

Anyway, are these the variables in the same units (bytes)?

BlackDex: In my opinion i think the current way of outputting this is correct because it limits the values returned to LibreNMS via the agent tool. And you do not have to change LibreNMS it self everytime an update is done.

But yes, the variables are in the same units. I have it running for a while now, and it works.

BlackDex:
I have signed the cla, didn't noticed that i had to do that, sorry.
Please merge this :).

PR #112, opened 2017-03-30
various misc fixes for the postfix poller

VVelox:
https://github.com/librenms/librenms/blob/master/includes/polling/applications/postfix.inc.php

~~Move $msefl to be before $sce as it is in the current poller.~~

Also a few misc. changes to better process a few items.

### DIFF
--- a/snmp/postfixdetailed
+++ b/snmp/postfixdetailed
@@ -142,7 +142,6 @@ my $recipientsC=0;
 my $recipienthdC=0;
 my $deferralcrC=0;
 my $deferralhidC=0;
-my $chrC=0;
 my $hcrnfqhC=0;
 my $sardnfC=0;
 my $sarnobuC=0;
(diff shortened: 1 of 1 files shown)


snmp/postfixdetailed
```diff
@@ -500,10 +507,10 @@ my $data=$received."\n".
 	$rardnf."\n".
 	$rarnfqa."\n".
 	$iuscp."\n".
+	$msefl."\n".
```
laf: Why is this moved further up? Doesn't it mean this will now be broken for existing users?

VVelox: Yes, because the recently committed PR broke it. This syncs it to that.

https://github.com/librenms/librenms/pull/6173

laf: It's much easier to fix in the main repo otherwise users have to re-deploy the application

VVelox: Unless we are pulling librenms/librenms#6173 completely that has already happened.

laf:
Main repo pr has been reverted so this is no longer needed

VVelox:
Okay moved $mself back to the end and done some other minor cleanup and corrections.

Nothing that breaks current installs.

PR #114, opened 2017-04-06
hddtemp: fix shell expansion for drive detection

putnam:
I have a machine with 58 disks. hddtemp only detected 26 of them.

Currently the hddtemp script is only able to find 26 drives due to the use of '?'. Switching it to the more aggressive '*' will pick up drives after sda-sdz, like sdaa, sdab, etc.

I tried to think of some cases where you wouldn't want the full expansion but was unable to think of any.

### DIFF
--- a/agent-local/hddtemp
+++ b/agent-local/hddtemp
@@ -11,6 +11,9 @@
 # the source code distribution for details.
 #
 # requires which, find, awk and sed
+#
+# optionally, install gnu parallel for a significant performance boost 
+# on machines with large numbers of drives.
 
 # Try to use lsblk if available. Otherwise, use find.
 if type lsblk >/dev/null 2>&1; then
@@ -23,7 +26,13 @@ hddtemp=`which hddtemp 2>/dev/null`
 
 if [ "${hddtemp}" != "" ]; then
 	if [ -x "${hddtemp}" ]; then
-		content=`${hddtemp} -w -q ${disks} 2>/dev/null | awk -F": " 'BEGIN{ ORS="" }{ print "|"$1"|"$2"|"$3"|";} ' | sed 's/[° ]C|/|C|/g' | sed 's/[° ]F|/|F|/g'`
+		if type parallel > /dev/null 2>&1; then
+			# When available, use GNU parallel for a significant performance boost. hddtemp runs serially(!)
+			output=`parallel ${hddtemp} -w -q ::: ${disks} 2>/dev/null`
+		else
+			output=`${hddtemp} -w -q ${disks} 2>/dev/null`
+		fi
+		content=`echo "$output" | awk -F": " 'BEGIN{ ORS="" }{ print "|"$1"|"$2"|"$3"|";} ' | sed 's/[° ]C|/|C|/g' | sed 's/[° ]F|/|F|/g'`
 		if [ "${content}" != "" ]; then
 			echo '<<<hddtemp>>>'
 			echo ${content}
--- a/agent-local/hddtemp
+++ b/agent-local/hddtemp
@@ -13,7 +13,7 @@
 # requires which, awk and sed
 
 # If disks are missing, they can be added here:
-disks="/dev/hd? /dev/sd?"
+disks="/dev/hd* /dev/sd*"
 
 hddtemp=`which hddtemp 2>/dev/null`
 
--- a/agent-local/hddtemp
+++ b/agent-local/hddtemp
@@ -11,9 +11,6 @@
 # the source code distribution for details.
 #
 # requires which, find, awk and sed
-#
-# optionally, install gnu parallel for a significant performance boost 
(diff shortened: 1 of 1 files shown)


exarv:
I don't think this is a correct patch, because now hddtemp also tries to retrieve the temperature from all partitions like /dev/sda1 /dev/sda2.

Maybe it's better to do the following:
disks="/dev/hd[a-z] /dev/hd[a-z][a-z] /dev/sd[a-z] /dev/sd[a-z][a-z]"

putnam:
You're totally right. Obvious miss on my part. I didn't expect hddtemp to
output anything for the partitions, but it sure does.

There must be a cleaner way. Is it safe to assume lsblk is on every modern
system? lsblk -d would effectively list all devices and no partitions.

putnam:
This, for example, would be effective:


```bash
disks=`lsblk -dnp|cut -d' ' -f1|tr '\n' ' '`
```

exarv:
I don't know if lsblk is available on all distro's but I think the lsblk solution is the best way to go.
Testing it on a server, it also gives back a nvme SSD. (although hddtemp does not support it unfortunately)

But... aren't we doing things to difficult ??
When I run hddtemp without any arguments, it already gives back all devices that is supports...
# hddtemp 
/dev/sda: SAMSUNG MZ7LM240HCGR-00005: 41°C
/dev/sdb: SAMSUNG MZ7LM240HCGR-00005: 42°C
/dev/sdc: SAMSUNG MZ7LM240HCGR-00005: 42°C
/dev/sdd: SAMSUNG MZ7LM240HCGR-00005: 40°C
/dev/sde: SEAGATE ST1000NM0023: 38°C
/dev/sdf: SEAGATE ST1000NM0023: 38°C
/dev/sdg: SEAGATE ST1000NM0023: 44°C
/dev/sdh: SEAGATE ST1000NM0023: 45°C
/dev/sdi: SEAGATE ST1000NM0023: 45°C
/dev/sdj: SEAGATE ST1000NM0023: 44°C
/dev/sdk: SEAGATE ST1000NM0023: 42°C
/dev/sdl: SEAGATE ST1000NM0023: 39°C
/dev/sdm: SEAGATE ST1000NM0023: 45°C
/dev/sdn: WDC WD4000FYYZ-01UL1B2: 51°C
/dev/sdo: WDC WD4000FYYZ-01UL1B2: 52°C
/dev/sdp: WDC WD4000FYYZ-01UL1B2: 49°C
/dev/sdq: WDC WD4000FYYZ-01UL1B2: 45°C
/dev/sdr: SAMSUNG MZ7LM240HCGR-00003: 44°C
/dev/sds: SAMSUNG MZ7LM240HCGR-00003: 42°C
/dev/sdt: INTEL SSDSC2BB240G6: 37°C
/dev/sdu: INTEL SSDSC2BB240G6: 39°C

So maybe the script can be made much simpeler :)

putnam:
I don't experience that behavior with hddtemp on my machine:

$ hddtemp
Too few arguments: you must specify one drive, at least.

I'm on Ubuntu 16.10 (yakkety) and this is the version in their repo:
hddtemp version 0.3-beta15

So I guess we can't have that luxury here if we want some amount of
compatibility.

exarv:
That's odd...

# hddtemp --version
hddtemp version 0.3-beta15

Running on CentOS 7

But than I would agree to use the lsblk

putnam:
Yes, it looks like your version has a custom patch added:

http://pkgs.fedoraproject.org/cgit/rpms/hddtemp.git/commit/?id=c8f5eb32e80749e7974513600866e8dc2609dd8e

putnam:
For what it's worth, here's how they did it:

```C
    if(glob("/dev/[hs]d[a-z]", 0, NULL, &diskglob) == 0) {
```

I still think blkid may be the "right" way going forward but certainly globbing like this will avoid the dependency.

exarv:
hmm. so they also don't see devices like sdaa :)
I also found a hddtemp-all.sh in the contrib dir of hddtemp.

That script is using:
df -l |cut -f 1 -d " " |grep /dev/ |sed -e 's/[[:digit:]]$//g' |sort -u |

But I still think that the lsblk solution is best.

exarv:
Maybe check if lsblk is available, if not use the glob variant (/dev/[hs]d[a-z] /dev/[hs]d[a-z][a-z] ?

putnam:
I like that idea. Adjusted code coming...

putnam:
I decided to use find instead of globbing. It's cleaner and more legible, included in every distro, and doesn't have to deal with ugly bash-isms with a series of multiple globs that may or may not match.

putnam:
I was working on something else and made a mess just now -- I can make a new pull request if/when this is approved if you'd like to keep it clean.

putnam:
(Creating a new pull request to clean up the mess)

PR #115, opened 2017-04-06
hddtemp: improve detection of drives

putnam:
NOTE: This is a continuation of a previously closed pull request (#114) because I'm bad at git.
---

Previously, this script was only able to find 26 drives (sda-sdz) due to the use of globbing.

A better strategy for detecting drives would be to use lsblk on systems that support it, failing over to globbing.

This patch adds support both for lsblk and a more comprehensive glob solution with find that will at least catch 26^2 drives.

### DIFF
--- a/agent-local/hddtemp
+++ b/agent-local/hddtemp
@@ -10,10 +10,14 @@
 # option) any later version.  Please see LICENSE.txt at the top level of
 # the source code distribution for details.
 #
-# requires which, awk and sed
+# requires which, find, awk and sed
 
-# If disks are missing, they can be added here:
-disks="/dev/hd? /dev/sd?"
+# Try to use lsblk if available. Otherwise, use find.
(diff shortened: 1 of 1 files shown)


exarv:
Look good, I think.

If I wanted to be a nitpicker, I would say to include tr into the 'requirement' comments.
And maybe mention lsblk and cut as optionals :)

putnam:
It seemed kind of silly that the script even enumerates the "requirements" as they're completely standard in every distro... even those that use busybox!

exarv:
Yeah, maybe you are right. I'm using 90% CentOS, so I was not sure if all those tools are the same in every distro.

laf:
I'm assuming this is backwards compatible?

putnam:
Yes, it is backward compatible. It detects for the presence of lsblk and
does not use it if it is not available.

PR #117, opened 2017-04-07
hddtemp: parallelize calls to hddtemp for performance

putnam:
This poll script runs hddtemp with a list of all drives as arguments and reads the output. hddtemp scans each drive's SMART status serially, which scales poorly with a large number of drives.

In lieu of a patch to the actual hddtemp project, optionally use GNU parallel when available to parallelize the call to hddtemp.

In testing a machine with 58 drives I went from a runtime of about 5 seconds per run to 0.5s, a performance improvement of 10x.

### DIFF
--- a/agent-local/hddtemp
+++ b/agent-local/hddtemp
@@ -11,6 +11,9 @@
 # the source code distribution for details.
 #
 # requires which, find, awk and sed
+#
+# optionally, install gnu parallel for a significant performance boost 
+# on machines with large numbers of drives.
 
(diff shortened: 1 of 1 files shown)


putnam:
@laf  @exarv I think mention bot might not have worked on this one for some reason.

As I was working on hddtemp the other day I realized it is really slow due to the serial nature of hddtemp's polling. This patch adds support for parallelizing the calls on systems with GNU parallel installed.

PR #119, opened 2017-04-20
Update munin

dragans:
Change line 3 cd command to full path, to make sure scripts works according to guidelines at

http://docs.librenms.org/Extensions/Applications/#munin


f0o:
That's the wrong path, it's supposed to be a different folder containing only the scripts you want to run. that directory is local in the same WD as the munin script on execution.

PR #120, opened 2017-04-22
Include missing SMART ids in the output

RedChops:


laf:
Does this change require any poller changes or is it fixing a bug?

RedChops:
Sorry I wasn't more specific. It's a small bug being fixed. The poller expects these fields to be returned, they're specified in the script higher up but aren't read from the smartctl output.

PR #122, opened 2017-04-23
snmp-extend: sdfsinfo

crcro:
bash script to export sdfscli --volume-info to snmp


laf:
I think php is a bit ott for this, can it not be done in bash or python?

PR #131, opened 2017-05-29
Added devuan support for os-updates.sh and removed code repitition

aldemira:
Since ubuntu and devuan are debian clones they could be merged in a single line.

### DIFF
--- a/snmp/os-updates.sh
+++ b/snmp/os-updates.sh
@@ -51,14 +51,7 @@ if [ -f /etc/os-release ]; then
                 else
 			echo "0";
 		fi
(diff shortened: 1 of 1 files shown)


laf:
I'm not a user of any of those - are they 100% compatible?

aldemira:
Yup, they all use apt-get, and the script uses standard parameters to calculate pending updates for debian alike. However, if you want we can keep this PR waiting for sometime to see if anyone objects. 
Btw, I don't have ubuntu handy but I've tested this with a devuan and debian system.

PR #133, opened 2017-05-31
C.H.I.P. power values

drid:
Get Voltage and Current values for AXP209 buses.
Get battery charge percentage.


drid:
Missing attribution.

PR #136, opened 2017-06-08
Strip all non-ASCII characters from hddtemp output

einarjh:
This works around garbage output from the hddtemp utility by stripping away everything but CR, LF and all characters from octal 40 to 176 (up until DEL). Fixes librenms/librenms#6800.


laf:
Thanks for the fix :)

PR #145, opened 2017-09-01
Handle (very) old RedHat version

rockyluke:


snmp/distro
```diff
@@ -28,6 +28,9 @@ elif [ "${OS}" = "Linux" ] ; then
       DIST="Mandriva"
       PSEUDONAME=`cat /etc/mandriva-release | sed s/.*\(// | sed s/\)//`
       REV=`cat /etc/mandriva-release | sed s/.*release\ // | sed s/\ .*//`
+    elif [ "${DIST}" = "Red" ]; then
+      DIST="RedHat"
+      IGNORE_LSB=1
```
laf: So this is the only real change as the `else` below should have set the dist to redhat already?

rockyluke: It's "Red Hat" with a space and we need to ignore LSB file, so I was thinking that's the best choice (but it's quite ugly I agree with you).

PR #148, opened 2017-09-19
fix munin agent

tomarch:
Without the full munin-scripts path, this script won't find munin file and return nothing.

### DIFF
--- a/agent-local/munin
+++ b/agent-local/munin
@@ -1,6 +1,6 @@
 # Lokale Einzelchecks
 export MUNIN_LIBDIR=/usr/share/munin
-if cd munin-scripts
+if cd $MUNIN_LIBDIR/munin-scripts
 then
   for skript in $(ls)
   do


laf:
Same as proposed here which has a comment in:

https://github.com/librenms/librenms-agent/pull/119

tomarch:
Ok it's the same, but request was refused for wrong reason.
/usr/share/munin/ is provided by the munin package
but all of the default munin plugins is in /usr/share/munin/plugins/
/usr/share/munin/munin-scripts/ is the folder created for the librenms agent. We have to place desired plugin in it.
So "cd /usr/share/munin/munin-scripts/" is OK, it will only graph the desired plugin and not all the default munin plugin.

f0o:
Hrm and I just wondered why my install broke.. this isnt backwards compatible as all scripts used to be relative to script's path.

Ah well... updating installs..

PR #150, opened 2017-10-10
Use command -v rather than absolute path

addp009:


tracernz:
Ahh, this is a better idea. :+1:

SourceDoctor:
nearly 2 years nothing happens
also completed outdated
running version returns also JSON

PR #152, opened 2017-11-29
Snmp scripts

xbeaudouin:
Added in the makefile install of all snmp scripts into distribution and updated debian changelog.


Makefile
```diff
@@ -4,8 +4,11 @@ install:
 	mkdir -p $(PREFIX)/usr/lib/check_mk_agent/plugins
 	mkdir -p $(PREFIX)/usr/lib/check_mk_agent/repo
 	mkdir -p $(PREFIX)/usr/lib/check_mk_agent/local
+	mkdir -p $(PREFIX)/usr/lib/librenms_agent
```
laf: Why these paths? They aren't documented anywhere, /usr/lib/check_mk_agent/local is where we tell people to install the scripts.

xbeaudouin: The scripts in /usr/lib/librenms_agent are snmp extensions. Since they are not the same as check_mk they have to be somewhere.
I don't like too much to move them in /usr/local/sbin or /usr/sbin since they can conflicts with some tools.

laf: We point snmp extensions to `/etc/snmp/` as per the current applications docs.

xbeaudouin: Well using /etc/snmp is somewhat strange, since /usr/lib/check_mk is used to store all others plugins.
Also saving scripts in /etc which is usualy for configuration files only, is IMHO not a good idea for debian / lsm purists.

murrant: @xbeaudouin but the snmp extend scripts cannot be used with check_mk, they don't print the required header.

PR #154, opened 2017-12-07
Small hack to incorporate Linux netfilter (nftables)

w41l:
Currently only support one nftables set. This should be determined automatically because on default installation fail2ban will make different of set for each jail: f2b-<setname> (ie. f2b-sshd, f2b-nginx, etc)

Screenshot (before and after adding nft support):
[https://s20.postimg.org/fb5vxcpfx/portal-fail2ban-librenms.png](https://s20.postimg.org/fb5vxcpfx/portal-fail2ban-librenms.png)

### DIFF
--- a/snmp/fail2ban
+++ b/snmp/fail2ban
@@ -9,6 +9,9 @@
 #make sure this path is correct
 my $f2bc="/usr/bin/env fail2ban-client";
 
+# Choose linux firewall type (nft or iptables)
+my $linuxFirewall="nft";
+
 #make sure this path is correct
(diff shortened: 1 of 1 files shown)


laf:
This now makes the default nft when it was iptables before so you need to switch that back around.

It's actually probably better to read that in via a config file as well so people can choose which one without editing the script.

w41l:
Yes, you are right. I forgot about the main config file. It has been ages since the last time I touch my config. I'm closing this PR for now.

PR #155, opened 2017-12-21
update the fail2ban stuff

VVelox:
This is for a shortly incoming update to the fail2ban polling.

### DIFF
--- a/snmp/fail2ban
+++ b/snmp/fail2ban
@@ -1,68 +1,103 @@
 #!/usr/bin/env perl
+# Author: Zane C. Bowers-Hadley <vvelox@vvelox.net>
 
-# Add this to your snmpd.conf as below.
-# extend fail2ban /etc/snmp/fail2ban
-#
-# Then add to your cron tab, if you wish to use caching.
-# */3 * * * * /etc/snmp/fail2ban.pl -u
+# https://docs.librenms.org/#Extensions/Applications/#fail2ban
+# See the above for additional information not documented in the POD below.
 
-#make sure this path is correct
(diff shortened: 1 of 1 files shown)


karlshea:
This update breaks on Ubuntu 16.04 when not running as root:

```
root@citywall:/etc/snmp# ./fail2ban
1
sshd 1
sshd-ddos 0
root@citywall:/etc/snmp# sudo -i -u librenms /etc/snmp/fail2ban
ERROR  Permission denied to socket: /var/run/fail2ban/fail2ban.sock, (you must be root)
Use of uninitialized value $jailsS in substitution (s///) at /etc/snmp/fail2ban line 105.
Use of uninitialized value $jailsS in substitution (s///) at /etc/snmp/fail2ban line 106.
Use of uninitialized value $jailsS in split at /etc/snmp/fail2ban line 107.
0
```

murrant:
I think this means you weren't fully collecting the data before.

Give the snmp user read access to the socket perhaps?

karlshea:
image

It was working until I pulled the latest librenms-agent, including both jails.

karlshea:
If I set the socket to world rwx it does continue to work, but that might need to be part of the app's installation instructions.

murrant:
Odd that there was any change as that was outside of the script.  The fail2ban-client is used for both scripts.

An update to the docs would be appreciated, but it would need to be more generally applicable.

karlshea:
It is odd, I don't really have an explanation. I half-heartedly looked through the diff and saw it was using iptables before too, but calling iptables as the snmp user doesn't work either.

VVelox:
This script required using sudo to to call it on some systems previously.

PR #156, opened 2017-12-21
ZFS SNMP agent :3 <3

VVelox:
Wooho!

Sorry about only having a FreeBSD one at this time. But it is ready to go.

### DIFF
--- a/snmp/zfs-freebsd
+++ b/snmp/zfs-freebsd
@@ -2,7 +2,7 @@
 
 =head1 DESCRIPTION
 
-Thgis is a SNMP extend for ZFS and FreeBSD for use with LibreNMS.
+This is a SNMP extend for ZFS and FreeBSD for use with LibreNMS.
 
 For more information, see L<https://docs.librenms.org/#Extensions/Applications/#zfs>.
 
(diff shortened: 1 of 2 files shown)


snmp/zfs-freebsd
```diff
@@ -0,0 +1,266 @@
...
+
+Pretty print the JSON.
+
+=head1 SNMPD SETUP EXAMPLES
+
+    extend zfs /etc/snmp/zfs-freebsd
+
+=cut
+
+#Copyright (c) 2017, Zane C. Bowers-Hadley
```
laf: We need a GPLv3 compatible license, I can't tell what license this is from the below.

VVelox: that is the standard 2 clause BSDL

VVelox: https://www.gnu.org/licenses/license-list.en.html#FreeBSD

That with "The views and conclusions contained in the software and documentation are those of the authors and should not be interpreted as representing official policies, either  expressed or implied, of the FreeBSD Project." removed from the end.

enilfodne:
Sorry in advance to discuss this in the PR, but i don't want to open an issue if its intended behavior.

I was wondering if the output length might be a problem?

I have 2 FreeBSD machines and the output from `zfs-freebsd` is ~1500 and ~1800 characters, respectively. However `NET-SNMP` times-out when the `zfs` extend is enabled on the second machine (~1800 chars), while it works fine on the first one (~1500).

Connection is UDP/1500MTU

PR #158, opened 2017-12-30
extend: powerdns-dnsdist

crcro:
snmp extend script needed for powerdns-dnsdist support

### DIFF
--- a/snmp/powerdns-dnsdist
+++ b/snmp/powerdns-dnsdist
@@ -2,16 +2,14 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-API_AUTH_USER=""
-API_AUTH_PASS=""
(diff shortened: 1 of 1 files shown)


murrant:
Why are you processing the data here?  Just dump it and process it in LibreNMS.

Also, the powerdns_recursor script allows direct api access.

crcro:
the api response only reports incremental values (not last values since last poll)

murrant:
Then use counter rrd type?

crcro:
i meant that you would get only an up line every pool (like uptime graph), not very useful, i think that queries per 5 minute graph is better that 100 queries 10 minutes ago, 192 queries 5 minutes ago, 203 queries now graph.

PR #160, opened 2018-01-18
Change Red Hat things

FTBZ:
Red hat detection is not working well.

- Added the flag to disable LSB
- Added the unused PSEUDONAME variable
- Removed Oracle from the Red hat section

Debug:
https://p.libren.ms/view/baafebda

⚠️ Please inspect the code in deep.

### DIFF
--- a/snmp/distro
+++ b/snmp/distro
@@ -51,7 +51,7 @@ elif [ "${OS}" = "Linux" ] ; then
   elif [ -f /etc/debian_version ] ; then
     DIST="Debian `cat /etc/debian_version`"
     REV=""
-    ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/  //g'`
+    ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
     if [ "${ID}" = "Raspbian" ] ; then
       DIST="Raspbian `cat /etc/debian_version`"
     fi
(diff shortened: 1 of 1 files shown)


snmp/distro
```diff
@@ -19,23 +19,25 @@ elif [ "${OS}" = "Linux" ] ; then
...
       DIST="CentOS"
     elif [ "${DIST}" = "Mandriva" ]; then
       DIST="Mandriva"
       PSEUDONAME=`cat /etc/mandriva-release | sed s/.*\(// | sed s/\)//`
       REV=`cat /etc/mandriva-release | sed s/.*release\ // | sed s/\ .*//`
-    elif [ -f /etc/oracle-release ]; then
-      DIST="Oracle"
     else
-      DIST="RedHat"
+      DIST="RedHat ${PSEUDONAME}"
```
laf: This will change the distro and potentially mean peoples alerts will trigger. It will either need a notification going out or this change reverting so it's just RedHat.

FTBZ: The script isn't working to match Red Hat at this time. With the script like the original, the LSB section always replaces the Red Hat section and the returned informations by LSB are not matched by LibreNMS.

The final $OSSTR is something like RedHatEnterpriseLinux and it makes it match nothing. The name must be RedHat<space>Something to match the OS detection.

laf: But is RedHat as the distro not enough? That's all we do for other distros like Ubuntu, CentOS, etc afaik

SourceDoctor:
@FTBZ @laf 
PR still actual or is it outdated?

SourceDoctor:
no action since 2 years. Doesn't seem to be relevant anymore

PR #164, opened 2018-02-25
Replace Disk Id with Disk Serial in S.M.A.R.T script

enilfodne:
The script currently uses the drive id as assigned by the OS. Those are not unique and can get shuffled around even without changing hardware; However, a bigger issue is encountered, when the user replaces a drive and the SMART data from the old drive gets mixed up with the new one, because the id was re-used.

The changes in this PR, replace the disk identifier with the disk's serial number (taken from `smartctl`). The user still specifies the disk identifiers on their system (e.g. `ada`, `sda`, etc.), but they are not used as identifiers in the output/graphs. This should allow for a much more precise monitoring and it should be consistent even if the hardware changes (for example, changing controller/raid cards or moving to a new chipset).

The script will fallback to disk identifier, if `smartctl` is unable to get a serial number from the disk.

### DIFF
--- a/snmp/smart
+++ b/snmp/smart
@@ -239,6 +239,7 @@ my $toReturn='';
 my $int=0;
 while ( defined($disks[$int]) ) {
 	my $disk=$disks[$int];
+	my $disk_sn=$disk;
 	my $output=`$smartctl -A /dev/$disk`;
 
 	my %IDs=( '5'=>'null',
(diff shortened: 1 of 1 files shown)


laf:
Thanks for the PR.

This is going to mean a loss of historical data for users I guess but only if they upgrade the script. Not really sure how we should handle this for users. Will merge anyway.

Issue #141, opened 2017-08-02
dhcp-status.sh: fixed paths


craig:
dhcp-status.sh contains lots of fixed paths, and thus does not work on debian.

laf:
You will be best off submitting a pull request with a fix, it's unlikely someone else will do this in reality.

PR #165, opened 2018-03-01
SNMPd shell extends update

enilfodne:
Changes:

- Remove variables as commands.
This makes the scripts unreadable and needlessly complicated. Reminding the user to update `PATH` seems like a much simpler solution.

- Add a simple guide how to setup each script
While this is already in the documentation, i think its helpful to have it in the file as well. Especially when editing over `ssh`.

- Rewrite NTP (`client`|`server`)
The main issue was that the previous scripts were relying on specific order of the output which might (did) change. This version uses `export` to create variables from the `ntpq` output and thus ensure the correct value in the correct order

- Rewrite UPS (`apc`|`nut`)
As with NTP, this rewrite uses variables to create the expected output. The `nut` and `apcaccess` output is being processed to make the variable name compatible with export.

- Fix DHCP lease status
The beginning of the string contained and empty space (`\x20`), which prevented the correct parsing of the output.

Not all scripts are being updated, feedback is appreciated as im not sure if some of the scripts are being readable/accessible enough.

### DIFF
--- a/snmp/dhcp-status.sh
+++ b/snmp/dhcp-status.sh
@@ -1,16 +1,21 @@
-#!/bin/bash
-################################################################
-# copy this script to somewhere like /opt and make chmod +x it #
-# edit your snmpd.conf add the below line and restart snmpd    #
-# extend dhcpstats /opt/dhcp-status.sh                         #
-################################################################ 
+#!/usr/bin/env bash
+
+# - Copy this script to somewhere (like /opt/dhcp-status.sh)
+# - Make it executable (chmod +x /opt/dhcp-status.sh)
+# - Add the following line to your snmpd.conf file
+#   extend dhcpstats /opt/dhcp-status.sh
+# - Restart snmpd
+#
+# Note: Change the path accordingly, if you're not using "/opt/dhcp-status.sh"
+
+# You need the following tools to be in your PATH env, adjust accordingly
+# - cat, grep, sed, sort, tr, wc
+PATH=$PATH
+
+# Leases location
 FILE_DHCP='/var/lib/dhcp/db/dhcpd.leases'
-BIN_CAT='/usr/bin/cat'
-BIN_GREP='/usr/bin/grep'
-BIN_TR='/usr/bin/tr'
-BIN_SED='/usr/bin/sed'
-BIN_SORT='/usr/bin/sort'
-BIN_WC='/usr/bin/wc'
+
+# Patterns
 DHCP_LEASES='^lease'
 DHCP_ACTIVE='^lease|binding state active'
 DHCP_EXPIRED='^lease|binding state expired'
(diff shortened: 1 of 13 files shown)


enilfodne:
I used `master` and it seems to be the latest version of the scripts. Can you point me to a newer version?

f0o:
So the removal of the variables instead of the hardcoded function calls were intentional?

enilfodne:
> In this case, I'd call a blocker on the issue altogether. The variables were introduced because some binaries may not be in $PATH and therefore it should be easily changeable.

I got that, but i think that allowing a user to change `PATH` in order to match their system is a much cleaner solution than declaring full paths for each executable. After all, that's what the shell does with the `PATH` variable - It will look for an executable with the specified name, through each directory.

tracernz:
Better to leave the command vars, and set them to just the executable name, so it will depend on PATH by default but user has the option to edit script and insert full path.

enilfodne:
But why? The current behavior adds unnecessary verbosity. It makes the script difficult to read, difficult to maintain and users still have issues with them (#141).

If your concern is with conflicting versions, e.g. (there's an old version in `/bin` and a newer one in `/usr/local/bin`), the user just have to pre-pend `/usr/local/bin` to `PATH`, there's no need to assign variables.

I'm not trying to impose anything, i looked through the repo history and the only reason for the var change appears to be the different paths on different distros/OS. Which is why i think changing `PATH` is a much cleaner option, than asking the user to change (potentially) every single executable.

paulgear:
I'm a big +1 on these changes; the originals were not good examples of bash style, and PATH is the right tool to solve different locations.

paulgear:
I had been meaning to make these changes myself for quite a while, but I just haven't had the brain space to devote to LibreNMS for the last year or so.

tracernz:
I like what #150 does.

paulgear:
Even #150 is unnecessary verbosity and poor shell style.

tracernz:
Assigning the paths to variables, or utilizing `command`? I can agree the former isn't the best. Using `command` at each of the call sites would be good IMO.

paulgear:
@tracernz Assigning variables for every command used.  It makes the script much harder to read and much longer than it needs to be.

tracernz:
Yeah, fair enough. You are correct that PATH is the right tool. In the snmp case you'd need to alter PATH wherever snmpd is invoked. 

So, if this is merged it would likely create merge conflicts for every other PR here... Will be a bit messy but worth it?

tracernz:
Thinking more about how to manipulate PATH for net-snmp extend scripts...
I think you have two options:
 - Set PATH for the whole snmpd process. For snmpd on a modern Linux distro that would mean copying /lib/systemd/system/snmpd.service to /etc/... and adding `Environment=PATH=$PATH:...` which is not too painful, and your maintenance burden has moved from librenms-agent scripts, to service declaration (packager may change flags to suit new versions etc. which you'd need to stay on top of). I guess it's not much different for SysV init type systems.
- Use a rc file + interactive shell, e.g.: `extend /bin/sh -i --rcfile /script/to/setup/PATH /opt/librenms-agent/snmp/script`

Is this correct?

enilfodne:
You can only realistically do the 3rd option as extends are looked as external programs and executed as such. Modifying the `snmpd` process to accommodate them is a recipe for trouble. Its simpler and self-contained to just edit the `PATH` on the script/program itself and keep the rest of the system untouched. 

The reason why i made this PR is because i can see 7 active PR that try do the same thing - change the variables or the way the variables derive the path to the respective program. The irony is not lost on me, but i thought that I'm using these scripts anyway, better share it and see if someone finds them useful.

laf:
Would prefer to keep this open for now

enilfodne:
There are too many open PR that basically, try to fix the same problem. I don't think this is going to be useful to someone, which is why i closed it in the first place. Plus, seeing an open PR for so long is often perceived as careless author. I do care and this is why i want to close this PR.

tracernz:
It would be nice to clean all the scripts up in one giant PR (like this one) and then start building/fixing on top of that again. Maybe setup travis-ci to run ShellCheck or some other linter over PRs to help stop them degrading after that.

PR #168, opened 2018-04-08
make using SN or device name selectable for SMART reporting

VVelox:
This fixes the breakage introduced in https://github.com/librenms/librenms-agent/pull/164 .

Unfortunately that commit breaks previous installs if people update to it. This makes it togglable in the config file, defaulting to the safe option of not using SN reporting.

The problem with SN based reporting is it tracks per drive instead of per slot. meaning you loose history each time you replace a drive.

### DIFF
--- a/snmp/smart
+++ b/snmp/smart
@@ -42,7 +42,7 @@ line with out a = or # are treated as a disk.
 
     #This is a comment
     cache=/var/cache/smart
-    smartctl=/usr/bin/env smartctl
+    smartctl=/usr/local/sbin/smartctl
     useSN=0
     ada0
     ada1
@@ -52,6 +52,7 @@ The variables are as below.
     cache = The path to the cache file to use. Default: /var/cache/smart
     smartctl = The path to use for smartctl. Default: /usr/bin/env smartctl
     useSN = If set to 1, it will use the disks SN for reporting instead of the device name.
+            1 is the default. 0 will use the device name.
 
 If you want to guess at the configuration, call it with -g and it will print out what it thinks
 it should be.    
@@ -68,7 +69,7 @@ use Getopt::Std;
 my $cache='/var/cache/smart';
 my $smartctl='/usr/bin/env smartctl';
 my @disks;
-my $useSN=0;
+my $useSN=1;
 
 $Getopt::Std::STANDARD_HELP_VERSION = 1;
 sub main::VERSION_MESSAGE {
--- a/snmp/smart
+++ b/snmp/smart
@@ -43,6 +43,7 @@ line with out a = or # are treated as a disk.
     #This is a comment
     cache=/var/cache/smart
     smartctl=/usr/bin/env smartctl
+    useSN=0
     ada0
     ada1
 
@@ -50,6 +51,7 @@ The variables are as below.
 
     cache = The path to the cache file to use. Default: /var/cache/smart
     smartctl = The path to use for smartctl. Default: /usr/bin/env smartctl
+    useSN = If set to 1, it will use the disks SN for reporting instead of the device name.
 
 If you want to guess at the configuration, call it with -g and it will print out what it thinks
 it should be.    
@@ -59,13 +61,14 @@ it should be.
 ##
 ## You should not need to touch anything below here.
 ##
-my $cache='/var/cache/smart';
-my $smartctl='/usr/bin/env smartctl';
-my @disks;
-
 use warnings;
 use strict;
 use Getopt::Std;
+
+my $cache='/var/cache/smart';
+my $smartctl='/usr/bin/env smartctl';
+my @disks;
+my $useSN=0;
 
 $Getopt::Std::STANDARD_HELP_VERSION = 1;
 sub main::VERSION_MESSAGE {
@@ -169,7 +172,7 @@ if ( defined( $opts{g} ) ){
 		$matches_int++;
 	}
 	
-	print 'smartctl='.$smartctl."\n".
+	print "useSN=0\n".'smartctl='.$smartctl."\n".
 		$cache.
 		join( "\n", keys(%found_disks) )."\n";
 	
@@ -209,6 +212,10 @@ while ( defined( $configA[$configA_int] ) ){
 		$smartctl=$val;
 	}
 
+	if ( $var eq 'useSN' ){
+		$useSN=$val;
+	}
+	
 	if ( !defined( $val ) ){
 		push(@disks, $var);
 	}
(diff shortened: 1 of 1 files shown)


VVelox:
o.0 Blarg? I have no clue what the CLA check is freaking out about. :(

VVelox:
This makes it selectable for those who care.

Besides having to having making this use SN instead of device just adds more work from a administrative stand point.

And yes, they do matter. It allows matter as they will tell you if more than a drive is buggy. They can also indicate a potentially buggy slot as well.

And then tracking historic info by drive serial number just becomes a massive issue as you have to actually track that all some place over time as well, which is freaking rare.

By serial number is and edge case.

enilfodne:
> defaulting to the safe option of not using SN reporting.

What is unsafe about it? Your rrd file are stil in `/opt/librenms/rrd`, you can still view them through the interface. History is kept intact.

> The problem with SN based reporting is it tracks per drive instead of per slot. meaning you loose history each time you replace a drive.

It does not track per slot, it tracks the drive SN, which should be unique. Plug that drive in another slot, or an HBA and the data continues. The idea behind it is that you care about the S.M.A.R.T of an individual drive, not the slot or the OS assignment.

> Besides having to having making this use SN instead of device just adds more work from a administrative stand point.

This is really outside the scope of LibreNMS.

> And yes, they do matter. It allows matter as they will tell you if more than a drive is buggy. They can also indicate a potentially buggy slot as well.

I think this is also outside the scope of LibreNMS. S.M.A.R.T data is per-drive information source and should only show data relevant to said drive. The conclusion of a bad port/slot should be left to the administrator. This app is tracking the drives output.

> And then tracking historic info by drive serial number just becomes a massive issue as you have to actually track that all some place over time as well, which is freaking rare.

You can create a new application allows you to define slots and join `rrd` from different drives. I think it makes no sense for a `S.M.A.R.T` app to care about the slot, position or how many drives were in the same slot previously.

> By serial number is and edge case.

Every storage appliance tracking `S.M.A.R.T` data, does it on a per-drive basis, because that's what `S.M.A.R.T` is. It makes no sense to care about the arbitrary assignment, slot number or order.

I would advice against this being the default option.

VVelox:
> What is unsafe about it? Your rrd file are stil in /opt/librenms/rrd, you can still view them through the interface. History is kept intact.

No it is not. A new RRD is created each time a drive is swapped. Creating a massive pain in the ass of dealing with it and tracking data over time.

> I think this is also outside the scope of LibreNMS. S.M.A.R.T data is per-drive information source and should only show data relevant to said drive. The conclusion of a bad port/slot should be left to the administrator. This app is tracking the drives output.

I wrote this app. Trust me when I say you are completely talking out of your ass on that here.

Also that is drek reasoning as you are intentionally arguing making it harder to track said data because it does not confirm to your idea of how it should work.

> You can create a new application allows you to define slots and join rrd from different drives. I think it makes no sense for a S.M.A.R.T app to care about the slot, position or how many drives were in the same slot previously.

Again you have cleary never done much sysadmin work before if you think such.

The entire purpose of this app is debugging and tracking drive issues. I don't care if a drive has been replace. I just care if the component in there currently is working.

Your suggestion is akin tracking the network graphs via MAC address.

> Every storage appliance tracking S.M.A.R.T data, does it on a per-drive basis, because that's what S.M.A.R.T is. It makes no sense to care about the arbitrary assignment, slot number or order.

Then clearly you've not done much server management. While it may not make sense for some appliances, if it is anything built on Dell or Supermicro gear I can guarantee you it makes complete sense.

SN is basically useless from a sysadmin standpoint.

Again you are missing the point of troubling shooting bad hardware.

enilfodne:
> No it is not. A new RRD is created each time a drive is swapped. Creating a massive pain in the ass of dealing with it and tracking data over time.

The `rrd` will be listed in the interface and you just have to click on them. Granted after a lot of replacements you will lose track of which drives was in which slot, but again, this is not about slot debugging or troubleshooting.

> I wrote this app. Trust me when I say you are completely talking out of your ass on that here.

Which im grateful for. You wrote, someone else improved it and we all continue to benefit from each other's work. That's how OSS works.

> The entire purpose of this app is debugging and tracking drive issues. I don't care if a drive has been replace. I just care if the component in there currently is working.

This is a monitoring application, not a debugger or troubleshooter. If you don't care about a single drive, this should not be named `S.M.A.R.T`. The previous solution was akin to caddy monitoring.

> Then clearly you've not done much server management. While it may not make sense for some appliances, if it is anything built on Dell or Supermicro gear I can guarantee you it makes complete sense.
>
> SN is basically useless from a sysadmin standpoint.
>
> Again you are missing the point of troubling shooting bad hardware.

I completely understand why you're doing this way and that it saves the administrator the hassle to look for the drive. However this is outside the scope of this application. LibreNMS monitors the drives and records the data. Its up to the administrator to interpret it.

Side-note: please refrain from comments about my past experience or expertise.

VVelox:
> The rrd will be listed in the interface and you just have to click on them. Granted after a lot of replacements you will lose track of which drives was in which slot, but again, this is not about slot debugging or troubleshooting.

It does all of the above. It monitors it for the current disk. 

Also no, the old data is not saved. If you replace a disk then you have a new SN, meaning the interface won't show it. Besides this just creates more hassle for every one.

> Which im grateful for. You wrote, someone else improved it and we all continue to benefit from each other's work. That's how OSS works.

Then stop acting like a complete ass hole and telling me what I wrote and why I wrote it.

And this is a regression. If said function is desired it should be toggle.

> This is a monitoring application, not a debugger or troubleshooter. If you don't care about a single drive, this should not be named S.M.A.R.T. The previous solution was akin to caddy monitoring.

Right. And it was monitoring the SMART data of the disk at that location. This is akin to displaying network device name by MAC address instead of device name then loosing data if it changes.

> However this is outside the scope of this application. LibreNMS monitors the drives and records the data.

Right and the change that was made makes it more problematic and annoying to do it.

> Side-note: please refrain from comments about my past experience or expertise.

Then at least show some level of experience in said field. And stop telling my why I wrote said application and what I wrote it for. That is just beyond insulting.

enilfodne:
I thin its pointless to debate any further. It seems you want this application to fit your workflow and environment. I just tried to make this more generic and applicable to more people.

I have no issue with this option being a toggle, my only issue is being the default option. I think at this point a maintainer should step in and decide what's best.

VVelox:
> I thin its pointless to debate any further. It seems you want this application to fit your workflow and environment. I just tried to make this more generic and applicable to more people.

Less generic and more tailored to subset large storage admins specifically currently instead of generic server stuff. 

> I have no issue with this option being a toggle, my only issue is being the default option. I think at this point a maintainer should step in and decide what's best.

You should of said so. =^.^=

I personally don't care either way as long as it is documented and toggle.

That said, sorry about being overly grumpy/pissy and jumping into this on a massive lack of sleep.

laf:
I'll be honest and say I'm not fussed which way we go with this. We have 3 people who obviously have an opinion on this. @VVelox @tracernz @enilfodne.

Please reply confirming which default you want and we will just go with the majority.

VVelox:
I'll be happy to make the change to the default. It defaults is good, as long as it can be toggled.

And again I just want to say sorry about jumping on this all on a lack of sleep and getting grumpy of it.

tracernz:
So, the core purpose for SMART is monitoring drives, not hardware/buses that drives are connected to. From that angle, tracking by SN absolutely makes sense. The data can however be used to monitor a few external things indirectly, which is VVelox's use-case. Thus I would so that tracking by SN should be default, with track by (ephemeral) name as an option. 

The downside of the second option is, if you don't populate the bus in order, your drives will be renamed at a later date (at least in Linux anyhow, that's why you mount by UUID/PARTUUID rather than e.g. /dev/sda2 in fstab) and the newer drives will in inherit the history of the older drives, while the older ones will start fresh. Additionally, when replacing a drive, it may not be clear in the history where that occurred.

VVelox:
> So, the core purpose for SMART is monitoring drives, not hardware/buses that drives are connected to. From that angle, tracking by SN absolutely makes sense. The data can however be used to monitor a few external things indirectly, which is VVelox's use-case. Thus I would so that tracking by SN should be default, with track by (ephemeral) name as an option.

Its true for any actual server monitoring case, not just mine. Seeing SMART data as being limited to only the drive is more of a lack of experience, IMHO. Its akin to seeing error data on a network interface as purely indicating a bad NIC than possible bad cable/fiber.

> The downside of the second option is, if you don't populate the bus in order, your drives will be renamed at a later date when you populate the lower numbered slots (at least in Linux anyhow. For example, that's why you mount by UUID/PARTUUID rather than e.g. /dev/sda2 in fstab) and the newer drives will in inherit the history of the older drives, while the older ones will start fresh. Additionally, when replacing a drive, it may not be clear in the history where that occurred.

This is only an issue if the kernel you are using does not nail stuff down statically, which FreeBSD does.

Also when it was replaced is sort of irrelevant to the purpose of the monitoring and data collection. Just like it is pointless to care about when a NIC is replaced. Think of this all as similar to the error info collected in regards to NICs. :)

That all said, I am fine with either as the default. =^.^=

VVelox:
Okay. Made the swap of it over to using SN by default. :)

laf:
We all happy now and we can merge this? @VVelox @tracernz @enilfodne.

tracernz:
LGTM 👍

Issue #166, opened 2018-03-14
zfs-freebsd: Can't run on FreeNAS


tracernz:
It requires a perl module that doesn't come with FreeNAS, and you can't really alter OS packages on FreeNAS because they get blown away each time you upgrade. What I did is port the script to python and only used the standard library.

Note shebang line specific to system... `/usr/bin/env python3` failed on FreeNAS so I haxed it. Needs fixed properly.
```python
#!/usr/local/bin/python3

import json
import subprocess

def percent(numerator, denominator, default=0):
	try:
		return numerator / denominator * 100
	except ZeroDivisionError:
		return default

def main(args):
	p = subprocess.run(['/sbin/sysctl', '-q', 'kstat.zfs', 'vfs.zfs'], stdout=subprocess.PIPE, universal_newlines=True)
	
	if p.returncode != 0:
		return p.returncode

	def chomp(line):
		bits = [b.strip() for b in line.split(':')]
		return bits[0], int(bits[1])
	stats = dict(chomp(l) for l in p.stdout.splitlines())
	if 'kstat.zfs.misc.arcstats.recycle_miss' not in stats:
		stats['kstat.zfs.misc.arcstats.recycle_miss'] = 0

	output = dict()

	# ARC misc
	output['deleted'] = stats['kstat.zfs.misc.arcstats.deleted']
	output['evict_skip'] = stats['kstat.zfs.misc.arcstats.evict_skip']
	output['mutex_skip'] = stats['kstat.zfs.misc.arcstats.mutex_miss']
	output['recycle_miss'] = stats['kstat.zfs.misc.arcstats.recycle_miss']

	# ARC size
	output['target_size_per'] = stats['kstat.zfs.misc.arcstats.c'] / stats['kstat.zfs.misc.arcstats.c_max'] * 100
	output['arc_size_per'] = stats['kstat.zfs.misc.arcstats.size'] / stats['kstat.zfs.misc.arcstats.c_max'] * 100
	output['target_size_arat'] = stats['kstat.zfs.misc.arcstats.c'] / stats['kstat.zfs.misc.arcstats.c_max']
	output['min_size_per'] = stats['kstat.zfs.misc.arcstats.c_min'] / stats['kstat.zfs.misc.arcstats.c_max'] * 100

	output['arc_size'] = stats['kstat.zfs.misc.arcstats.size']
	output['target_size_max'] = stats['kstat.zfs.misc.arcstats.c_max']
[...]

PR #170, opened 2018-04-08
add zfs support for linux

Serphentas:
This is just to bring support for ZFS probing on Linux devices where SPL is used with ZoL.

It's basically a Python rewrite of [zfs-freebsd](https://github.com/librenms/librenms-agent/blob/master/snmp/zfs-freebsd) using some parts of the code written by @tracernz in #166.


laf:
Thanks for this @Serphentas. I assume this outputs the same as the freebsd version?

Can you also update the docs here to show a linux version is available along with the instructions to install: https://github.com/librenms/librenms/blob/master/doc/Extensions/Applications.md#zfs

Serphentas:
@laf You're welcome. As for the output, I've made sure all JSON keys read in the [poller code](https://github.com/librenms/librenms/blob/master/includes/polling/applications/zfs.inc.php) are present and correct.

Sure, will update docs now.

laf:
@Serphentas All good to go on this PR?

Serphentas:
@laf Yes you can go ahead and merge. I just committed a quick fix.

Also, to note is that the `-p` parameter for the `zpool list` is not available on all ZFS versions, or at least not with version `0.6.5.6-0ubuntu16`.

As for the documentation, see https://github.com/librenms/librenms/pull/8546.

PR #174, opened 2018-04-24
convert all the NTP stuff to JSON

VVelox:
This requires https://github.com/librenms/librenms/pull/8571 and is for https://github.com/librenms/librenms/pull/8608 .

Also converted this to regular sh instead of bash, so it will work on more systems with less dependencies.

Has been tested as working on DD-WRT and FreeBSD.

### DIFF
--- a/snmp/ntp-client.sh
+++ b/snmp/ntp-client.sh
@@ -26,23 +26,23 @@ VERSION=1
 #error and errorString are hardcoded as if the above fails bad json will be generated
 RAW=`$BIN_NTPQ -c rv | $BIN_GREP jitter | $BIN_SED 's/[[:alpha:]=,_]/ /g'`
 if [ $NTPQV = "p11" ]; then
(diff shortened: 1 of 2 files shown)


enilfodne:
I think using JSON in the output is a really messy option:

* JSON is not really needed. The admin controls the sender and the receiver and LibreNMS usually keeps them compatible, so using kv system is unnecessary
* JSON is too verbose. For example, the ZFS agent does not work on my machine (with 2 pools), when using UDP, because the output exceeded the maximum packet size
* Its one more dependency or hackish printing in order to produce the output

Honestly, there's really no point in changing any of the agents, as no changes are discussed or accepted. You can see how many PR are still open, just for NTP agent.

VVelox:
@enilfodne If you are having an issue with it and UDP, it sounds more like you are having some sort of network issue that needs to be investigated.. That should be fine. Here it produces a minimum 1849 byte reply for me, which easily exceeds my MTU of 1500.

VVelox:
As to moving to JSON, it allows easier stuff to be done in the future.

PR #175, opened 2018-04-29
Changed to use proxy instead of extend in snmpd.

jzadir:
Its way simpler to use proxy and direct the queries to squid.

### DIFF
--- a/snmp/squid
+++ b/snmp/squid
@@ -1,74 +1,15 @@
 #!/bin/sh
 
-# Add this to snmpd.conf as below.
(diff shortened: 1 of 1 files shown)


SourceDoctor:
still useful?
As here described [https://docs.librenms.org/Extensions/Applications/#squid](https://docs.librenms.org/Extensions/Applications/#squid) there is no Agent script needed,
just configuration of squid config and snmpd config

SourceDoctor:
script not needed, so it's removed.
Documentation is in Application Docs

PR #182, opened 2018-07-14
Update distro for CoreOS support

crazy-max:


laf:
This isn't needed with your PR to add support directly into LibreNMS

crazy-max:
@laf Oh ok, so `distro` is kind of deprecated ?

laf:
No, it's just one or the other. Most linux distros can't be detected over snmp so we use the distro script.

PR #183, opened 2018-07-17
Asterisk Script

TheGreatDoc:
Asterisk App support.
- Channels
- Calls
- Total SIP Peers
- Monitored Online
- Monitored Offline
- Unmonitored Online
- Unmonitored Offline

### DIFF
--- a/snmp/asterisk
+++ b/snmp/asterisk
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+ASCLI=/usr/sbin/asterisk
+
+if [ -f $ASCLI ];
+then
+    $ASCLI -rx "core show uptime" > /dev/null
+    if [ $? -ne 0 ]; then
+        # Asterisk not running, silently exit.
+        exit 0
+    fi
+
+    echo "<<<asterisk>>>"
+    $ASCLI -rx "core show channels" | awk '/active calls/ { print "Calls=" $1 } /active channels/ { print "Channels=" $1}'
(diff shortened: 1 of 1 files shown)


mattie47:
@TheGreatDoc  - How'd you possibly get around getting the script to run as a different user? 

For me, Asterisk runs as "asterisk" meanwhile I think a lot of people historically ran it as root. 

So, whenever I access the asterisk console, I use "sudo asterisk -rvvvv" rather than switching to the asterisk user. 

As a result, the script for snmp would run as user "snmp", meaning a walk would give:

```
iso.3.6.1.4.1.8072.1.3.2.4.1.2.8.97.115.116.101.114.105.115.107.1 = STRING: "Unable to connect to remote asterisk (does /var/run/asterisk/asterisk.ctl exist?)"
```

A "workaround" to get the script to run, was change the permissions in ```/etc/asterisk/asterisk.conf```


```diff
; Changing the following lines may compromise your security.
-;[files]
+[files]
-;astctlpermissions = 0660
+astctlpermissions = 7777
;astctlowner = root
;astctlgroup = apache
;astctl = asterisk.ctl
```

But I'm curious if you had a better method of connecting to the asterisk console as the snmp user for the script...

Thanks,

Matt

TheGreatDoc:
@mattie47 
In my case, asterisk runs as asterisk user and snmpd runs as root as default.

In your case, that you runs snmpd as normal user I suggest adding the snmp user to the astctlgroup that is apache in your case.

Hope thats help!

mattie47:
Thanks for that @TheGreatDoc

PR #184, opened 2018-07-19
Correct error zpool command

Cormoran96:
Zpool list option is P Uppercase not lowercase

### DIFF
--- a/snmp/zfs-linux
+++ b/snmp/zfs-linux
@@ -95,7 +95,7 @@ def main(args):
     PREFETCH_METADATA_MISSES_PERCENT = DEMAND_METADATA_MISSES / ARC_MISSES * 100
 
     # pools
-    proc = subprocess.run(['/sbin/zpool', 'list', '-pH'], stdout=subprocess.PIPE, universal_newlines=True)
+    proc = subprocess.run(['/sbin/zpool', 'list', '-PH'], stdout=subprocess.PIPE, universal_newlines=True)
     if proc.returncode != 0:
         return proc.returncode
 


laf:
this seems to vary :(

```
[root@nas2] ~# zpool list -PH
invalid option 'P'
usage:
	list [-Hpv] [-o property[,...]] [-T d|u] [pool] ... [interval [count]]
```

Cormoran96:
```bash
root@miraz:[~]#` lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
root@miraz:[~]# zpool list -pH
invalid option 'p'
usage:
	list [-gHLPv] [-o property[,...]] [-T d|u] [pool] ... [interval [count]]
```

So bad news 😢

laf:
If the exit code for that is non-zero then check that and run the original command (or vice versa)

Zmegolaz:
My zpool has both. :) 
`list [-gHLpPv] [-o property[,...]] [-T d|u] [pool] ... [interval [count]]`

From the manual:
```
             -p      Display numbers in parsable (exact) values.
             -P      Display full paths for vdevs instead of only the last component of the path. This can be used in conjunction with the -L -flag.
```
Some check that the output is what's expected seems to be needed. @Cormoran96 , what switch is used in your version to display exact numbers?

Cormoran96:
It is
```
zpool list [-T d | u] [-HgLPv] [-o props[,...]] [pool] ... [interval[count]]
           -P          Display full paths for vdevs instead of only the last component of the path.  This can be used in conjunction with the -L flag.
```
```
zpool get [-p] "all" | property[,...] pool ...
           -p    Display numbers in parseable (exact) values.


```

Zmegolaz:
`zpool get -p` is different from `zpool list -p`. `list` show a list of pools, while `get` show properties for the pools. This script want the list of pools. What does your `zpool list -H` look like?

Cormoran96:
it look like this:

```
root@miraz:[~]# zpool list -H
data	1.81T	83.1G	1.73T	-	2%	4%	1.00x	ONLINE	-
```

Zmegolaz:
I basically see two ways forward:

1. See if you run an old version of the zfsutils-linux package (or what it's called in your dist, whichever provides the `zpool` command) and update it.
2. Modify this script to check if the -p flag exists, otherwise calculate what the suffixes `k`, `M`, and so on means.

Any other ideas?

Cormoran96:
Moved to  #213

PR #185, opened 2018-07-25
Fix Debian detection on Proxmox - lsb_release binary doesn't exist

dsgagi:
IF condition for Debian systems also checks if lsb-release package is installed on the system.

Proxmox doesn't have it installed by default, for example.

### DIFF
--- a/snmp/distro
+++ b/snmp/distro
@@ -46,7 +46,7 @@ elif [ "${OS}" = "Linux" ] ; then
     DIST="Devuan `cat /etc/devuan_version`"
     REV=""
 
-  elif [ -f /etc/debian_version ] ; then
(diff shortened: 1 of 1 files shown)


dsgagi:
This could also work with
```
if [ -f /usr/bin/lsb_release ] ; then
    ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
```

Let me know which one would you prefer :)

ndum:
Hi,
Any news on this PR yet? For Proxmox it would be really advantageous. 

Thanks!

dsgagi:
Something weird happened with this PR. Couldn't rebase it to the master branch.

Will create a new PR for change.

PR #193, opened 2018-09-17
PhotonOS distro detection

vowywowy:
Detection before `/etc/os-release` since that is present yet missing the build number.

Not many people run PhotonOS so here are some screenshots:
Before:
image
After:
image


vowywowy:
> I believe I just merged another PR that uses awk, can you update this branch?

I think I got the one you're talking about. I just put the changes from https://github.com/librenms/librenms-agent/pull/194 underneath the `awk` check.

murrant:
Please submit an os icon to the main repo :)

PR #196, opened 2018-10-03
ntp-client data correction

sparknsh:
NTP was not displaying data right for linux servers. It was putting the frequency data into the offset data. This was giving bad graphs in the UI. Tested the correction on both RHEL and Debian based operating systems and all passes.

Remove the .sh to simplify for configuration management orchestration scripts.


sparknsh:
After working with my changes I noticed that NTP stat output positions still would change over time. 
My new changes search the stat output for the variables we need and get it's true value and not assume its where it should be.

PR #204, opened 2018-10-18
Fix package manager detection

sparknsh:


sparknsh:
@murrant I have adjusted the script to still use the env variables.

sparknsh:
@murrant or @laf Can you merge this in?

snmp/osupdate
```diff
@@ -28,39 +28,39 @@ CMD_PACMAN='-Sup'
...
+elif [ -f '/etc/arch-release' ]; then
     # Arch
     UPDATES=`$BIN_PACMAN $CMD_PACMAN | $BIN_WC $CMD_WC`
     if [ $UPDATES -ge 1 ]; then
         echo $(($UPDATES-1));
     else
         echo "0";
     fi
-elif [ -f "$BIN_YUM" ]; then
+elif [ -f '/etc/redhat-release' ]; then
```
murrant: This doesn't cover centos.

sparknsh: It sure does. 

`cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)`

murrant: Let me be more generic.  This doesn't include some random distro that uses yum and doesn't have /etc/redhat-release.  Replace yum and /etc/redhat-release with any combo.

sparknsh: Made the changes

murrant:
Getting warmer here, but I think it is more portable to test the return code of command -v instead of the output.

murrant:
Thanks for working on this @sparknsh 

After a bit of testing, I _think_ this is what you want:
```
if command -v ls &>/dev/null ; then echo 'yes'; fi
```

sparknsh:
@murrant can you merge this in?

murrant:
Sure

PR #205, opened 2018-10-18
NGINX error fix due to syntax

sparknsh:


sparknsh:
@murrant or @laf Can you merge this in?

PR #206, opened 2018-10-21
Correct indenting in nginx script

xudonax:
There is an error in the current `snmp/nginx` script where the indenting is not correct. This should be 8 spaces instead of the current 2. This pull request fixes that.


sparknsh:
PR #205 already fixes the issue.

PR #209, opened 2018-10-29
Use which to solve executables path

crazy-max:
I've got many errors while using some agent scripts :

```
/etc/snmp/dhcp-status.sh: line 25: /usr/bin/cat: No such file or directory
/etc/snmp/dhcp-status.sh: line 25: /usr/bin/grep: No such file or directory
0
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/sed: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
0
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/sed: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
0
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/sed: No such file or directory
/etc/snmp/dhcp-status.sh: line 29: /usr/bin/grep: No such file or directory
```

We should use which to solve executables path.

### DIFF
--- a/snmp/asterisk
+++ b/snmp/asterisk
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ASCLI=/usr/sbin/asterisk
+ASCLI=`which asterisk`
 
 if [ -f $ASCLI ];
 then
--- a/snmp/dhcp-status.sh
+++ b/snmp/dhcp-status.sh
@@ -5,12 +5,12 @@
 # extend dhcpstats /opt/dhcp-status.sh                         #
 ################################################################ 
 FILE_DHCP='/var/lib/dhcp/db/dhcpd.leases'
-BIN_CAT='/usr/bin/cat'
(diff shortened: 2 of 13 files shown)


sparknsh:
which is not a path on all systems. That is why /usr/bin/env is used in most scripts.

sparknsh:
@crazy-max While you're doing these changes be sure to update the file name to match the snmp extend name just like what i did with osupdate, ntp-client, phpfpmsp, and nginx

crazy-max:
@sparknsh Ok you mean #199 ?

PR #211, opened 2018-11-21
Fix: InnoDB stat support for MariaDB v10+

slashdoom:
An attempt to fix issue in forum thread:
https://community.librenms.org/t/mysql-innodb-monitoring-on-nan/4608/11


tristanvs:
Testing as we speak, will keep you posted. Thanks for the patch!

PR #214, opened 2018-12-12
Fix Debian detection on Proxmox - lsb_release binary doesn't exist

dsgagi:
IF condition for Debian systems also checks if lsb-release package is installed on the system.

Proxmox doesn't have it installed by default, for example.

This could also work with

```
elif [ -f /etc/debian_version ] ; then
    DIST="Debian `cat /etc/debian_version`"
    REV=""
    if [ -f /usr/bin/lsb_release ] ; then
        ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
    fi
    if [ "${ID}" = "Raspbian" ] ; then
        DIST="Raspbian `cat /etc/debian_version`"
    fi
```

Let me know which one would you prefer :)

### DIFF
--- a/snmp/distro
+++ b/snmp/distro
@@ -54,10 +54,10 @@ elif [ "${OS}" = "Linux" ] ; then
     DIST="Debian `cat /etc/debian_version`"
     REV=""
     if [ -f /usr/bin/lsb_release ] ; then
-        ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
+      ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
     fi
     if [ "${ID}" = "Raspbian" ] ; then
-        DIST="Raspbian `cat /etc/debian_version`"
+      DIST="Raspbian `cat /etc/debian_version`"
     fi
 
   elif [ -f /etc/gentoo-release ] ; then
--- a/snmp/distro
+++ b/snmp/distro
@@ -50,12 +50,14 @@ elif [ "${OS}" = "Linux" ] ; then
     DIST="Devuan `cat /etc/devuan_version`"
     REV=""
 
-  elif [ -f /etc/debian_version -a -f /usr/bin/lsb_release ] ; then
+  elif [ -f /etc/debian_version ] ; then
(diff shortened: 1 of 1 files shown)


avarins:
Just what I need, can you do this mergen?
@crazy-max @murrant @laf @f0o

murrant:
Need people to test to make sure A. It works and B it doesn't break anything.

sbaerlocher:
The patch works for me, it doesn't give me any more error messages.

I would prefer the example of @dsgagi 

```bash
elif [ -f /etc/debian_version ] ; then
    DIST="Debian `cat /etc/debian_version`"
    REV=""
    if [ -f /usr/bin/lsb_release ] ; then
        ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
    fi
    if [ "${ID}" = "Raspbian" ] ; then
        DIST="Raspbian `cat /etc/debian_version`"
    fi
```

output:

```bash
Debian 9.6
```

This one has the following that I don't prefer so much.

```bash
  elif [ -f /etc/debian_version -a -f /usr/bin/lsb_release ] ; then
    DIST="Debian `cat /etc/debian_version`"
    REV=""
    ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
    if [ "${ID}" = "Raspbian" ] ; then
      DIST="Raspbian `cat /etc/debian_version`"
    fi
```

output:

```bash
Debian GNU/Linux 9
```

ndum:
Works for me too, a merge would be really nice. Thanks

crazy-max:
@dsgagi I'm also in favor of this impl :

```
elif [ -f /etc/debian_version ] ; then
    DIST="Debian `cat /etc/debian_version`"
    REV=""
    if [ -f /usr/bin/lsb_release ] ; then
        ID=`lsb_release -i | awk -F ':' '{print $2}' | sed 's/	//g'`
    fi
    if [ "${ID}" = "Raspbian" ] ; then
        DIST="Raspbian `cat /etc/debian_version`"
    fi
```

avarins:
I find the suggestion of @crazy-max good

dsgagi:
Done :)

PR #217, opened 2019-01-07
Added gpsd script for SNMP Extend

mikecentola:
Script to get GPSD data into SNMP extend. Matching PR for librenms coming next.

### DIFF
--- a/snmp/gpsd
+++ b/snmp/gpsd
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2019 Mike Centola <mcentola@appliedengdesign.com>
+#
+# Please make sure the paths below are correct.
+# Alternatively you can put them in $0.conf, meaning if you've named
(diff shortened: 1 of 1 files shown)


snmp/gpsd
```diff
@@ -0,0 +1,45 @@
...
+# this script gpsd.sh then it must go in gpsd.sh.conf .
+#
+#
+################################################################
+# Don't change anything unless you know what are you doing     #
+################################################################
+
+BIN_GPIPE='/usr/bin/env gpspipe'
+BIN_GREP='/usr/bin/env grep'
+BIN_PYTHON='/usr/bin/env python'
```
murrant: Since you are using python, why not just make this a python script?

mikecentola: I had thought about just using python to do the whole thing but GPSPIPE outputs individual JSON docs.

murrant: Whatever you prefer.  I just really dislike bash it is so hard to read.

PR #225, opened 2019-03-18
portactivity fixes

VVelox:
- whitespace cleanup
- always use -n to avoid a hyphen in Parse::Netstat
- make it actually work with -n and as a result on Linux and not just FreeBSD


VVelox:
This is backwards compatible. Nothing changes argument wise. Just it works now on Linux and some possible bugs in Parse::Netstat are avoided.

PR #226, opened 2019-03-19
SMART monitoring update

VVelox:
Now capable of monitoring RAID. Arbitrary flags may be added for each device configured as well as arbitrary names configured as well.

Below is a example monitoring a 3 disk 3ware RAID. twl0,0-2 are the returned device names and everything after the space is what is fed to smart.

```
    twl0,0 /dev/twl0 -d 3ware,0
    twl0,1 /dev/twl0 -d 3ware,1
    twl0,2 /dev/twl0 -d 3ware,2
```

The config stuff is also backwards compatible.

Also now only uses --scan-open only with the -g option. Also excludes known bad returns such as CD drives and SAS backplanes.

### DIFF
--- a/snmp/smart
+++ b/snmp/smart
@@ -1,5 +1,5 @@
 #!/usr/bin/env perl
-#Copyright (c) 2017, Zane C. Bowers-Hadley
+#Copyright (c) 2019, Zane C. Bowers-Hadley
 #All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


evanrich:
This stilll seems to aggregate tw10,0-2 into "tw10" in the UI

PR #228, opened 2019-04-16
Added SAS Drive Wrapping

Munzy:
The current SMART agent doesn't work at all with SAS drives. To resolve this I made a wrapper that injects "Elements in Grown Defect List:" into smart "5 Reallocated_Sector_Ct". This seems to be the most logical placement for these errors. Further I have taken drive temps from the SAS output and injected them into "194 Temperature_Celsius". 

I have currently tested this on a couple of SAS drives I have laying around. However, more testing may be required to verify that it works on all SAS drives.

I hope this helps.

### DIFF
--- a/snmp/smart
+++ b/snmp/smart
@@ -339,7 +339,41 @@ foreach my $line ( @disks ){
 				$IDs{$id}=$temp;
 			}
 
+
+            
+
 		}
+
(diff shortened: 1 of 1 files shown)


VVelox:
Howdy! A quick test here of it and it looks good.

PR #231, opened 2019-05-05
dhcp-status - seperate binary path configuration from code

SourceDoctor:
seperate code from path definition for binaries

**dhcp-status.conf**
```
FILE_DHCP='/var/lib/dhcp/dhcpd.leases'
BIN_CAT='/bin/cat'
BIN_GREP='/bin/grep'
BIN_TR='/usr/bin/tr'
BIN_SED='/bin/sed'
BIN_SORT='/usr/bin/sort'
BIN_WC='/usr/bin/wc'
```

often binary pathes differs from default settings, so only config file has to modified

### DIFF
--- a/snmp/dhcp-status.sh
+++ b/snmp/dhcp-status.sh
@@ -4,6 +4,7 @@
 # edit your snmpd.conf add the below line and restart snmpd    #
 # extend dhcpstats /opt/dhcp-status.sh                         #
 ################################################################ 
+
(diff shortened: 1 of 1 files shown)


SourceDoctor:
@VVelox i've changed it to an optional set

SourceDoctor:
this application is fixed ;)

VVelox:
Looks good!

PR #232, opened 2019-05-09
fix dhcp-status configfile declaration

SourceDoctor:
Configfile will not be found if snmpd is calling extend, so absolute path declaration is required.


SourceDoctor:
mmh strange
i signed license/cla, but it stays in pending state

PR #234, opened 2019-05-23
enhance smart to show power_on_hours also

SourceDoctor:
harddisc "power on hours" will be published with this patch


VVelox:
Looks good and works here. :)

PR #235, opened 2019-05-27
fixing configfile call

SourceDoctor:
the last pull request was incomplete sorry. With this pull request should work now as expected

fixes: #141, #129


SourceDoctor:
after this fix the configfile works as expected

SourceDoctor:
with this patch Issue #141 will be resolved. Also Pull Request #129 will be finally obsolet

PR #237, opened 2019-06-12
Allow a non-root user for bind

dlangille:
You won't want all of this pull request.  You'll want to drop the two changes to env.

And probably set bind_user to ''

But the rest is good.

### DIFF
--- a/snmp/bind
+++ b/snmp/bind
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/local/bin/perl
 #Copyright (c) 2017, Zane C. Bowers-Hadley
 #All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


VVelox:
Howdy!

This will work on FreeBSD, but will fail on Linux.

/usr/bin/env $command is also used as that grabs it via default as various OS and distros put it in different locations. Although now that I get thinking on it, it could just be shortened to $command. 

That said the best thing for non-root use for with the extend scripts is sudo. :) That way we don't need to add support for it to all the scripts and make sure the chosen method works across lots  of different OSes.

dlangille:
Yes, I know it will fail in Linux. I said "You'll want to drop the two changes to env" to make it clear this wasn't for all. The better approach would have been to not included in the patch at all.  My apologies for the extra work involved.

I like the `sudo` approach, but I didn't want to look at changing how the extend scripts are invoked.

VVelox:
Howdy!

It is no problem. Yeah, really the best thing is sudo. Just looking at updating the app docs to mention that at the top. :)

PR #238, opened 2019-06-19
Added opcahce.php file for webserver (SNMP)

SniperVegeta:


snmp/opcache
```diff
@@ -0,0 +1,22 @@
+<?php
```
Jellyfrog: This should be
```suggestion
#!/usr/bin/env php
<?php
```

snmp/opcache
```diff
@@ -0,0 +1,22 @@
...
+
+# This is the status script that goes with the Cacti Templates for opcache via SNMP
+# It can be accessed as a snmpd extension with something like:
+#       extend phpopcache /usr/bin/curl --silent http://localhost/opcache.php
```
Jellyfrog: This comment is wrong, it should be a path to the script directly.
Like:
```suggestion
#       copy this script to /etc/snmp/
#       chmod +x /etc/snmp/osupdate  
#       edit your snmpd.conf and include: extend phpopcache /etc/snmp/opcache
```

snmp/opcache
```diff
@@ -0,0 +1,22 @@
...
+# NOTE you should put an htaccess file on it for protection
+# need to be surved true the webserver for correct stats / CLI = other stats
+
+header('Content-Type: text/plain');
```
Jellyfrog: can be removed

SniperVegeta:
Well, i try that but, the CLI got other stats then true the webserver so thats why i curl to the php script inside the webserver....

Via the cli its not working....

SourceDoctor:
what is this for? which Application uses this?

PR #241, opened 2019-06-30
Correct directory check (really a dev / file), and wildcard given name

arrmo:
Correct directory check in mdadm - /dev/md* is not a directory, and also wildcard is needed.

Yell if you have any questions!

### DIFF
--- a/snmp/mdadm
+++ b/snmp/mdadm
@@ -14,7 +14,7 @@ ERROR_STRING=""
 
 OUTPUT_DATA='['\
 
-if [ -d /dev/md ] ; then
+if [ -e /dev/md* ] ; then
     for RAID in /sys/block/md* ; do
 
         # ignore arrays with no slaves


SourceDoctor:
thats true /dev/md* is not a directory.

But /dev/md is a directory which will be created and used by mdadm.

Also it's not necessary to call a mdadm device like this /dev/md*.
So checking for directory /dev/md is the most save way.

arrmo:
That makes sense - but I admit, without the changes it doens't work on my system (Ubuntu) ... :-(. Thoughts?

Thanks!

arrmo:
As for the CLA - signed that a long time ago. Hmmm.

SourceDoctor:
@arrmo
I tested an Ubuntu installation (18.10) with mdadm, and there also will be created an /dev/md directory

arrmo:
It could be because I'm on 19.04? Or because I have LVM? Just thinking out loud, not sure why exactly ... :-(. FYI, here is the directory,
```
/dev$ ls md*
md0
```

SourceDoctor:
there is no difference between Versions

If you have freshly created an mdadm array. did you run "update-initramfs -u" so array will be detected while booting?
is the array listed in /etc/mdadm/mdadm.conf ?

arrmo:
Hi,

Not a freshly created array - to be honest, it's been around for a very long time. And yes, it is detected during boot (info from dmesg below).
```
[    4.816514] md/raid1:md0: active with 2 out of 2 mirrors
[    4.816528] md0: detected capacity change from 0 to 2000263380992
[    7.445706] EXT4-fs (md0): mounted filesystem with ordered data mode. Opts: (null)
```
And yes, it's in /etc/mdadm/mdadm.conf (line below).
```
ARRAY /dev/md0 uuid=986b6934:c5c185f8:1115ad9f:e101fb33
```
Thoughts?

Thanks!

SourceDoctor:
very strange. When you'd take a look into mdadm manpage section "Auto assembly" you will see it writes symbolic link to it's block devices into /dev/md/ 
I also don't know why it didn't create this directory on your machine. But you can create it yourself, it's just a directory with symbolic links to it's block devices with it's name:

mkdir /dev/md
cd /dev/md
ln -s /dev/md0 0
ln -s /dev/md1 1
ln -s /dev/my_custom_array_name my_custom_array_name

arrmo:
Agreed - confused here too. It is odd, not sure if it's because of updates over time, or something else.

But if I have to create manual symbolic links ... I assume others may have the same issue? Perhaps a dumb question (likely is ... LOL), but - why not point directly to /dev/md* for the check? I may be missing why this is seen as safer.

Thanks!

SourceDoctor:
the reason is easy. you can customize your array names.
md0 ... mdN is default
it's also possible to name your array as you wish on creation ... so all arrays can only detected via their links in /dev/md/

Example with custom array name: 
mdadm --create /dev/my_custom_named_array --level=1 --raid-devices=2 /dev/sda /dev/sdb

arrmo:
OK, makes sense - thanks! Have to figure out how to make this work on my system then, as I'm not a big fan of manually making changes (like symlinks) - often find they cause me grief later ... :-(. Have to think about what to do.

Thanks again.

arrmo:
Closing, will just make local mods. Thanks for the help!

PR #242, opened 2019-06-30
mdadm - name independent array detection

SourceDoctor:
this patch modifies the mdadm script 
so names of mdadm devices can have every name now


SourceDoctor:
this patch get the arrayname given on creation for naming it in librenms, as it should

VVelox:
Looks good.

PR #243, opened 2019-07-01
Split SMART power on hours on h too, not only space

Zmegolaz:
Some drives report their `power on hours` (ID 9) as 24789h+02m+03.960s instead of integer hours. We only want the hours, so we split the string at `h` and use the first part.

### DIFF
--- a/snmp/smart
+++ b/snmp/smart
@@ -319,7 +319,7 @@ foreach my $line ( @disks ){
 
 			# 9, power on hours
 			if ( $id == 9 ) {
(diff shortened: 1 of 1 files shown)


VVelox:
Strange. Looks good and works here.

I am curious what drives are you seeing that with?

Zmegolaz:
Intel 530, 120 GB:
```
Model Family:     Intel 53x and Pro 2500 Series SSDs
Device Model:     INTEL SSDSC2BW120A4
```

PR #245, opened 2019-07-03
Merge nginx scripts

tuxis-ie:
I think we should prefer to use single scripts. This way, the nginx script is not depending on python2 or python3.


SourceDoctor:
@tuxis-ie in view on Merge conflicts, and only support for 
Python3 on Python Scripts, is your PR still relevant?

murrant:
Feel free to reopen

PR #247, opened 2019-07-10
add zfs support for freebsd use python 3

bestlong:
copy from #166.


VVelox:
Awesome. Looks like it works here.

PR #254, opened 2019-10-10
Updated named.stats location to be compatible with AppArmor

cppmonkey:
After upgrading a Debian Stretch server to Buster, there was a problem with the BIND service due to AppArmor preventing the service from writing to /var/run/named/

Perhaps we could ask the maintainer to add this location to the profile, but looking at the current profile included with bind9. I've updated the _bind_ snmp script to work with previous default and the new default.

dpkg -S /etc/apparmor.d/usr.sbin.named
bind9: /etc/apparmor.d/usr.sbin.named

https://downloads.isc.org/isc/bind9/cur/9.14/doc/arm/Bv9ARM.pdf
page 80 - statistics-file
Suggests named.stats would be in the servers current dir (/etc/bind/?) but AppArmor only grants read access

The named apparmor profile lists /var/lib/bind/ and /var/cache/bind/ as having rw access
  ```
# /etc/bind should be read-only for bind
  # /var/lib/bind is for dynamically updated zone (and journal) files.
  # /var/cache/bind is for slave/stub data, since we're not the origin of it.
  # See /usr/share/doc/bind9/README.Debian.gz
  /etc/bind/** r,
  /var/lib/bind/** rw,
  /var/lib/bind/ rw,
  /var/cache/bind/** lrw,
  /var/cache/bind/ rw,
```

### DIFF
--- a/snmp/bind
+++ b/snmp/bind
@@ -39,7 +39,7 @@ The variables are as below.
 
     rndc = The path to rndc. Default: /usr/bin/env rndc
     call_rndc = A 0/1 boolean on weather to call rndc stats. Suggest to set to 0 if using netdata. Default: 1
-    stats_file = The path to the named stats file. Default: /var/run/named/stats
+    stats_file = The path to the named stats file. Default: /var/cache/bind/named.stats
     agent = A 0/1 boolean for if this is being used as a LibreNMS agent or not. Default: 0
     zero_stats = A 0/1 boolean for if the stats file should be zeroed first. Default: 0 (1 if guessed)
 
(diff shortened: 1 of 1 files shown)


VVelox:
Hmm....  likely would be best to include a array with common locations and then check for its existence.

SourceDoctor:
@cppmonkey 
what about including an optional configfile so everybody could overwrite the existing pathes.
should be the best and most compatible way.
As example take a look in [dhcp-stats.sh](https://github.com/librenms/librenms-agent/blob/master/snmp/dhcp-status.sh)

cppmonkey:
Fixed by PR #290

PR #255, opened 2019-10-14
Added CPU Details

Munzy:
I have long wanted a way to monitor steal time on my VMs. I currently use a lot of VMs on hosting providers such as Linode, Digital Ocean, et cetera. These providers sometimes over-allocate a node to the point of becoming a problem. One way to indicate this issue is CPU Steal climbing beyond a certain percentage. 

https://scoutapm.com/blog/understanding-cpu-steal-time-when-should-you-be-worried

The goal is to allow LibreNMS to track users, system, idle, io wait, and steal numbers so that we can make better decisions on how to improve performance or mitigate performance issues on a VM.

### DIFF
--- a/snmp/detailedcpu.sh
+++ b/snmp/detailedcpu.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-# This is to get Detailed CPU stats into LibreNMS. 
-# The goal being to track Steal and IOWait in shared virtual enviornements. 
+# This is to get Detailed CPU stats into LibreNMS.
+# The goal being to track Steal and IOWait in shared virtual enviornements.
 
 # Please make sure you have tail, vmstat, and awk installed.
-# Currently modern kernels beyond 2.6.11 should be supported. 
+# Currently modern kernels beyond 2.6.11 should be supported.
 
 # By Munzy https://github.com/Munzy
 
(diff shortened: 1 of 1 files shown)


Munzy:
```
./detailedcpu.sh
0
0
100
0
0
```

murrant:
New applications should output json format with labeled fields.

Munzy:
> New applications should output json format with labeled fields.

Alright. I'll see how I want to refactor it.

snmp/detailedcpu.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
Jellyfrog:
```suggestion
#!/usr/bin/env bash
```

snmp/detailedcpu.sh
```diff
@@ -0,0 +1,31 @@
...
+# vmstat | tail -n 1 |  awk '{print $13, $14, $15, $16, $17}'
+# 0 0 100 0 0
+
+# $13 = User
+# $14 = System
+# $15 = Idle
+# $16 = IO Wait
+# $17 = CPU Steal
+
+VMSTAT=`vmstat 1 2 | tail -n 1`
```
Jellyfrog: This will run vmstat 2 times with 1 second interval

SourceDoctor:
rewritten to use same data via UCD MIB SNMP, which are usually transfered, also.

PR #256, opened 2019-10-18
Version fix for NTP SNMP

sparknsh:
The version variable was not pulled right on RedHat based distros. This fixes the issue.


Jellyfrog:
Whats the output on redhat?
Also this is a bit easier;
```bash
$BIN_NTPQ --version | awk 'match($0, /ntpd (.)/, a) {print a[1]}' 
```

sparknsh:
Before the fix it was (ntpq 4.2.6p5) now its (4) The full version was causing issues and not sending the data. 

That will not output anything. 
I have done my testing on el deb and arch. All passed.

PR #262, opened 2019-12-24
zfs-freebsd - Repair percentage calculation

dlangille:
When running this script, I get:

Useless use of multiplication (*) in void context at ./zfs-freebsd line 163


PipoCanaja:
Hi @dlangille 
Could you check this version ? Should give no error and the correct result now.

SourceDoctor:
@dlangille @PipoCanaja 
something new here?

dlangille:
Sorry, I have been preoccupied by PGCon and BSDCan.

PR #266, opened 2020-01-13
Cache package manager responses to reduce load

LukaszMoskala:
Hi,

I modified osupdate script a little to only execute package manager every 5 hours.
This is usefull on older hardware, where executing this script every 5 minutes caused significant resources usage.

Maybe this will be usefull for someone.


LukaszMoskala:
Commit made with wrong email

PR #271, opened 2020-01-22
Smart: fix useSN, selftest when disk is defined as a full argument to smartctl (arg after space)

fbourqui:
was resulting in /dev/dev/.... parameter to smartctl -l and -i


VVelox:
Awesome. Looks good, thanks!

PR #290, opened 2020-04-18
Change stats file location

AltiUP:
The stats file should not be placed in /var/run/named but in /var/cache/bind.
In run there will be permissions problems.

### DIFF
--- a/snmp/bind
+++ b/snmp/bind
@@ -39,7 +39,7 @@ The variables are as below.
 
     rndc = The path to rndc. Default: /usr/bin/env rndc
     call_rndc = A 0/1 boolean on weather to call rndc stats. Suggest to set to 0 if using netdata. Default: 1
-    stats_file = The path to the named stats file. Default: /var/run/named/stats
+    stats_file = The path to the named stats file. Default: /var/cache/bind/stats
     agent = A 0/1 boolean for if this is being used as a LibreNMS agent or not. Default: 0
     zero_stats = A 0/1 boolean for if the stats file should be zeroed first. Default: 0 (1 if guessed)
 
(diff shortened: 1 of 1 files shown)


SourceDoctor:
just change path of stats file is not such good.
statistics-file in bind config can be set as wished.

To make it more flexible it would be better to keep it as is
and overwrite configuration with an optional configuration file
So everybody can position stats file as prefered

AltiUP:
> just change path of stats file is not such good.
> statistics-file in bind config can be set as wished.
> 
> To make it more flexible it would be better to keep it as is
> and overwrite configuration with an optional configuration file
> So everybody can position stats file as prefered

Yes, of course. 
But I think we still need to remove the mentions to the locations in /var/run. 
It's misleading and bad practice.

SourceDoctor:
DNS statistics are nothing which needs to be cached.
They are saved there, but will never read from DNS.
So path is not such wrong.

Also there exists no official default path for statistics file,
so everybody can save it on prefered path.
If you'd like to have it on another path,
best way would be to enhance script with overwriting value with content from an optional configfile.

AltiUP:
Yes indeed, no official folder of the stats.

I just wanted to point out, that putting it in /var/run is a bad idea. 
For example, on Debian at reboot the permissions will be overwritten and therefore create an error when running the script.

SourceDoctor:
never had such problem on Debian with my bind9
just enhance that script with loading an optional configfile and everybody can handle it as wanted :-)
if you change the path directly in file some agents may stop fetching data

PipoCanaja:
Agreed with @AltiUP , this location is a nightmare for rights, on ubuntu and debian I had problems getting it right. RNDC itself can be painful to setup at first ... :) 
So any improvement here would be good. The best would be to get rid of this file but it seems that there is no way to get the stats in stdout instead of in a file.

Issue #278, opened 2020-03-04
librenms mdadm not working 


napaster:
Good day. Tell me how to use the module to monitor mdadm. For some reason, it doesn’t work for me, it produces some kind of incorrect data. More precisely, the script that you offer to download mdadm does not work at all.
My system: 
CentOS Linux release 7.7.1908 (Core)

> Installed Packages
Name        : mdadm
Arch        : x86_64
Version     : 4.1
Release     : 1.el7
Size        : 1.0 M
Repo        : installed
From repo   : base
Summary     : The mdadm program controls Linux md devices (software RAID arrays)
URL         : http://www.kernel.org/pub/linux/utils/raid/mdadm/
License     : GPLv2+
Description : The mdadm program is used to create, manage, and monitor Linux MD (software
            : RAID) devices.  As such, it provides similar functionality to the raidtools
            : package.  However, mdadm is a single program, and it can perform
            : almost all functions without a configuration file, though a configuration
            : file can be used to help with some common tasks.

> [root@store etc]# cat /etc/mdadm.conf 
ARRAY /dev/md0 metadata=1.2 spares=2 name=store.armada.loc:0 UUID=24097801:93f434e5:f9c9c442:5b14bafa
ARRAY /dev/md1 metadata=1.2 spares=2 name=store.armada.loc:1 UUID=efd14dc1:ac09359d:2cea9389:853e2811
ARRAY /dev/md2 metadata=1.2 spares=2 name=store.armada.loc:2 UUID=6f1b1e72:a3320daf:ea0c9116:9b022848

When I try to just execute the mdadm downloaded, I get the following output:

> [root@store snmp]# ./mdadm 
/etc/mdadm.conf: line 1: ARRAY: command not found
/etc/mdadm.conf: line 2: ARRAY: command not found
/etc/mdadm.conf: line 3: ARRAY: command not found
/etc/mdadm.conf: line 4: MAILADDR: command not found
{"data":],"error":"0","errorString":"","version":"1"}

What am I doing wrong ?

yo1frenchtoast:
Hello,

There was any issue with the check of /dev/md blocks presence, I made a PR to fix this

Also, you seem to use /etc/mdadm.conf as CONFIGFILE in mdadm script
I don't think it was the goal of this conf file, whatever the CONFIGFILE variable is not used anywhere in the script

Works for me now : 
```
root@openmediavault:/etc/snmp# ./mdadm 
[...]

PR #295, opened 2020-05-05
Fix bash check from directory to block type

yo1frenchtoast:
https://github.com/librenms/librenms-agent/issues/278

### DIFF
--- a/snmp/mdadm
+++ b/snmp/mdadm
@@ -17,8 +17,8 @@ ERROR_STRING=""
 OUTPUT_DATA='['
 
 # use 'ls' command to check if md blocks exist
-if ls /dev/md* 1> /dev/null 2>&1 ; then
-    for ARRAY_BLOCKDEVICE in $(ls -1 /dev/md*) ; do
+if $LS /dev/md?* 1> /dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


snmp/mdadm
```diff
@@ -14,10 +14,11 @@ VERSION=1
 ERROR_CODE=0
 ERROR_STRING=""
 
-OUTPUT_DATA='['\
+OUTPUT_DATA='['
 
-if [ -d /dev/md ] ; then
-    for ARRAY_BLOCKDEVICE in $(ls -1 /dev/md/*) ; do
+# use 'ls' command to check if md blocks exist
+if ls /dev/md* 1> /dev/null 2>&1 ; then
```
SourceDoctor, line 20: ls should be used from variable, as all other commands

SourceDoctor:
on running for test i got this:
```
[19:34h][root@testsys][:/root/teststuff]$ ./test.sh
/bin/ls: access on '/sys/block/md:/slaves' not possible: File or Directory not found
/bin/ls: access on '/sys/block/0/slaves' not possible: File or Directory not found
/bin/ls: access on '/sys/block/1/slaves' not possible: File or Directory not found
/bin/ls: access on '/sys/block/2/slaves' not possible: File or Directory not found
....
```
This shouldn't happen

PR #314, opened 2020-06-20
Scripts for Openwrt

arrmo:
Scripts to provide needed Wireless Sensor information back to LibreNMS - "agent" of sorts.


arrmo:
Hi,

Just checking - anything I need to do / add here? No issues if that's the case, just not sure.

Thanks!

PR #315, opened 2020-06-26
Use sh in ups-nut.sh shebang

moisseev:
since the `ups-nut.sh` script is not using any bash specific syntax.
This change removes unneeded dependency on bash.


SourceDoctor:
what's the improvement of this PR?

moisseev:
Portability, POSIX compliance. Not all OSes come with preinstalled bash.

PR #322, opened 2020-07-17
Fix missing compatibility to Pi-hole V5 API (minimal changes)

efelon:
To minimize the impact of future API changes in terms of data order or additional data being appended, required elements are selected directly with the "jq" tool in the expected order.

In theory this change should be backwards compatible to the previous version of the Pi-hole API, but I don't have access to an older version (V4 API) to test it.

### DIFF
--- a/snmp/pi-hole
+++ b/snmp/pi-hole
@@ -71,10 +71,10 @@ debug() {
 exportdata() {
 	# domains_being_blocked / dns_query_total / ads_blocked_today / ads_percentage_today
 	# unique_domains / queries_forwarded / queries_cached
-	GET_STATS=$(curl -s $API_URL$URL_READ_ONLY | jq '.[]')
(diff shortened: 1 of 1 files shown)


SourceDoctor:
is it maybe possible to get the API Version which is used and then switch command for it to ensure old versions stay functional?

elherr:
Thanks for updating the script!  Added changes from master and your commit on my system with Pi-hole 5.1 (latest) and works well.  Current API is version 3.
$ curl -s http://localhost/admin/api.php/?version | jq
{
  "version": 3
}

efelon:
@SourceDoctor I think it should also work with the previous version of the API. But, as I wrote in the PR, I don't have a Pi-hole 4 installation around. Someone confirming my assumption would be really great.

And then of course if it is not working I can add a API version check as suggested by @elherr.

elherr:
@efelon  I'd hope no one with LibreNMS installed is using the old Pi-hole API at this point.   If they are using old version of Pi-hole, I would have to guess they are not actively updating LibreNMS either.

PR #326, opened 2020-08-13
Rewrite Apache SNMP Agent to Python3

hanserasmus:
Thanks to @murrant this file is now python3 compatible.

### DIFF
--- a/snmp/apache-stats.py
+++ b/snmp/apache-stats.py
@@ -19,6 +19,7 @@
 #
 import os
 import time
+import urllib.request
 
 cachetime = 30
 cachefile = '/var/cache/librenms/apache-snmp'
@@ -31,10 +32,8 @@ if os.path.isfile(cachefile) and (time.time() - os.stat(cachefile)[8]) < cacheti
     data = f.read()
     f.close()
 else:
-    # Grab the status URL (fresh data), needs package python-urlgrabber
-    from urlgrabber import urlread
-    # "data" is UTF string, need to decode.
-    data = urlread('http://localhost/server-status?auto', user_agent='SNMP Apache Stats').decode()
+    # Grab the status URL (fresh data), needs package urllib3
+    data = urllib.request.urlopen("http://localhost/server-status?auto").read().decode('UTF-8')
     # Write file
     f = open(cachefile+'.TMP.'+str(os.getpid()), 'w')
     f.write(data)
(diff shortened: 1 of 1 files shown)


snmp/apache-stats.py
```diff
@@ -18,72 +18,68 @@
 #
 #
 
-CACHETIME = 30
-CACHEFILE = '/var/cache/librenms/apache-snmp'
+CACHETIME=30
```
SourceDoctor, line 21: it's python not bash
please do space between operators
-> PEP8 compatibilty

snmp/apache-stats.py
```diff
@@ -18,72 +18,68 @@
...
-if os.path.isfile ( CACHEFILE ) \
-	and ( time.time() - os.stat ( CACHEFILE )[8] ) < CACHETIME:
-	# use cached data
-	f = open ( CACHEFILE, 'r' )
-	data = f.read()
-	f.close()
+if os.path.isfile(CACHEFILE) \
+        and(time.time() - os.stat(CACHEFILE)[8]) < CACHETIME:
+        # use cached data
+        f=open(CACHEFILE, 'r')
```
SourceDoctor, line 30: also here

snmp/apache-stats.py
```diff
@@ -18,72 +18,68 @@
...
-	from urlgrabber import urlread
-	data = urlread ( 'http://localhost/server-status?auto',
-					user_agent = 'SNMP Apache Stats' )
-	# write file
-	f = open ( CACHEFILE+'.TMP.'+`os.getpid()`, 'w' )
-	f.write ( data )
-	f.close()
-	os.rename ( CACHEFILE+'.TMP.'+`os.getpid()`, CACHEFILE )
+        # Grab the status URL (fresh data), needs package python-urlgrabber
+        from urlgrabber import urlread
```
SourceDoctor, line 35: imports should be alway on top of the file

snmp/apache-stats.py
```diff
@@ -18,72 +18,68 @@
...
 ]
 for param in dataorder:
-	try:
-		print params[param]
-#                print param
-	except:	# not all Apache's have all stats
-		print 'U'
+    try:
+        print(params[param])
+    except: # not all Apache's have all stats
```
SourceDoctor, line 80: i think  `except KeyError:`  is better to catch only not existing key - errors

snmp/apache-stats.py
hanserasmus: Sorry, will do.

snmp/apache-stats.py
hanserasmus: You mean change:
`except:  # not....`

to
`except KeyError:  # not...`
?

PR #327, opened 2020-08-25
Add support for passing a password to the redis script

tobias-richter:
This PR adds password/auth support to redis monitoring
Example:
```
extend redis /etc/snmp/snmpd.d/redis.py -a "SOME_PASSWORD"
```


SourceDoctor:
what about an optional configfile which will be placed in /etc/snmp directory instead of script arguments?

PR #328, opened 2020-09-02
Update distro to support BeagleBoard

arrmo:
Add support for BeagleBoard (distro) - checks BeagleBoard signature, if existing (in /etc/dogtag). Example of that file,
```
cat /etc/dogtag
BeagleBoard.org Debian Buster IoT TIDL Image 2020-04-06
```


SourceDoctor:
LGTM

PR #340, opened 2020-12-10
Bugfix freeradius.sh

mpikzink:
Fix a syntax error. 

BTW I would prefer to consequently use no extension (some have):
"
Executables should have no extension (strongly preferred) or a .sh extension. Libraries must have a .sh extension and should not be executable.

It is not necessary to know what language a program is written in when executing it and shell doesn’t require an extension so we prefer not to use one for executables
" from 
https://google.github.io/styleguide/shellguide.html


SourceDoctor (approved):
good point

PR #341, opened 2020-12-16
cleaned up and improved freeradius monitor efficiency

thedaveking:
Changed freeradius.sh from making many grep calls to a single sed call. This reduced system calls from >11k to <3k.

I'm bummed I can't sign the CLA tho, because I have seen the other project's source code due to having subscribed at my previous job.

Anyone is welcome to use my code. None of it is influenced by that other project in any way.


thedaveking:
I'm closing this because I want to submit a better solution without the separate sed file

PR #344, opened 2021-01-18
Fix STDERR output corrupting json output in mdadm app

khobbits:
When checking if arrays have slaves, the mdadm script, does an  ls/$LS of the device to see if it exists.
This $LS throws an error to STDERR if it does not match.
This output is caught by snmp and corrupts the json output


murrant:
Thanks!

PR #351, opened 2021-02-11
add DHCP stats to PiHole

Chewie9999:
Added code to export the maximum DHCP leases available, and the current number of DHCP leases.
If DHCP is not enabled in Pi-Hole "0" is returned for both values.

Note, extra config is required when implementing this - the path to the Pi-Hole variables, and the path to the DHCP leases file.
If this is a standard install of Pi-Hole, this should not need changing.

An added dependency of the `calc` command is required.
This can be installed by issuing a `apt-get install calc` on the pi.

This script was tested on PiHole v5

Changed PHP files are also required on the librenms server. This will be in a separate pull request.

### DIFF
--- a/snmp/pi-hole
+++ b/snmp/pi-hole
@@ -8,11 +8,17 @@ API_AUTH_KEY=""
 API_URL="localhost/admin/api.php"
 URL_READ_ONLY="?summaryRaw"
 URL_QUERY_TYPE="?getQueryTypes&auth="
+PICONFIGFILE='/etc/pihole/setupVars.conf'
+DHCPLEASEFILE='/etc/pihole/dhcp.leases'
 
 if [ -f $CONFIGFILE ]; then
     . $CONFIGFILE
 fi
 
+# read in pi-hole variables for DHCP range
+if [ -f $PICONFIGFILE ]; then
+    . $PICONFIGFILE
+fi
 
 #/ Description: BASH script to get Pi-hole stats
 #/ Examples: ./pi-hole-stats.sh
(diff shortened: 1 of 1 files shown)


Chewie9999:
LibreNMS server push request : https://github.com/librenms/librenms/pull/12525

Jellyfrog:
You should be able to do the math in bash, iirc `**` for `^`

Chewie9999:
> You should be able to do the math in bash, iirc `**` for `^`

Thank you.
You are correct. I will resubmit the code without the `calc` dependency

Jellyfrog (approved):

si458:
sorry to tag onto this, `DHCP_ACTIVE` can generate an error if its not defined
my example: i run pi-hole in a docker container so the docker-host connects to the container ip to get stats,
so it doesnt know anything about the config file or how to get it,
only way around this is to set `DHCP_ACTIVE='false'` inside the pi-hole script itself

Chewie9999:
> sorry to tag onto this, `DHCP_ACTIVE` can generate an error if its not defined

> /etc/snmp/pi-hole: line 101: DHCP_ACTIVE: unbound variable
> root@dockerserver:~#
> ```
> 
> my example: i run pi-hole in a docker container so the docker-host connects to the container ip itself to get stats,
> so it doesnt know anything about the config file or how to read it,
> only way around this is to set `DHCP_ACTIVE='false'` inside the pi-hole script itself

Thanks @si458, I thought it was always there. I will check my code

PR #352, opened 2021-02-20
gpsd python error

kanokc:
gpsd script occasionally results in python error.

cause: the expected sentence from GPS unit on each update has more than 10 lines, therefore, python didn't find the expected wording on some update and result in a python error
correction: increase the number of sentence to get from gpspipe from 10 to 20 lines


SourceDoctor (approved):
Thx

PR #355, opened 2021-03-13
Cleanup some code

Jellyfrog:
Preparation for adding some linters

PR #362, opened 2021-03-23
Minor script reverts (osupdate, phpfpm)

jepke:
I needed to revert those changes to get the proper output which is handled by Librenms

I don't have any other then debian to test. But have more apps to add and test.

### DIFF
--- a/snmp/osupdate
+++ b/snmp/osupdate
@@ -58,8 +58,8 @@ elif command -v pacman &>/dev/null ; then
     fi
 elif command -v yum &>/dev/null ; then
     # CentOS / Redhat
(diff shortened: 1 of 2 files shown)


jepke:
Kinda expected linter to complain about this lol

Jellyfrog:
@jepke need to fix the failing test first

jepke:
@Jellyfrog this fixes the SNMP extend scripts that broke in #355 and #356 i honestly have no idea how to fix the lint issues all i know is that when i reverted the changes made last week all that i use work again.

I was able to test apt-get and yum for the osupdate script both broke when you applied the linter fixes in 355 and 356. I think its broken for the rest as well but unable to test.

PR #373, opened 2021-08-07
Added OS Security updates (for YUM only so far)

wimg:
Adds OS Security updates, useful when managing tons of servers to avoid notification fatigue caused by non-stop updates of packages.
This way, security updates can be placed in a rule separately from standard os updates.

Submitted the librenms repo part in https://github.com/librenms/librenms/pull/13117


SourceDoctor:
why not enhance osupdate plugin with this?
creating another OS Update Plugin makes no sense

wimg:
> why not enhance osupdate plugin with this?
> creating another OS Update Plugin makes no sense

I agree that might make more sense, but not all distros seem to have a reliable way of listing only security updates (apt-based for example seem to have an issue with this) and so modifying the entire osupdate plugin seemed a bit overkill.

Honestly I don't know librenms well enough to get started on that either.

SourceDoctor:
osupdate Script returns a Number.
fastest way would be to enhance this to a comma seperated list
and second field stays "Null" if no Sequrity Update Count is possible.

I personaly would prefer a line of this value in existing OS Update graph.
So you can see how many Update are open, and how many of them are Security Updates

take a look into smart Script to see an example

snmp/osupdate
```diff
@@ -39,7 +40,7 @@ if command -v zypper &>/dev/null ; then
     if [ "$UPDATES" -ge 2 ]; then
         echo $(($UPDATES-2));
```
SourceDoctor, line 41: what about this echos?

snmp/osupdate
```diff
@@ -39,7 +40,7 @@ if command -v zypper &>/dev/null ; then
...
     else
-        echo "0";
+        echo "0,0";
```
SourceDoctor, line 43: if a "section" does not support this feature ,
return 0 is a bad practice. "Null" is recommended

snmp/osupdate
wimg: I'm sorry, what do you mean ?

wimg:
Not really sure why it was doing :
echo $(($UPDATES-1));
in many places. With yum it's definitely not correct (unless you have some kind of problem with your setup), so I changed that to not do the -1.

wimg:
Of course this is a problem now :
image

Any pointers on how to fix that ?

wimg:
2 things that are not working :

- The RRD is giving an error because it suddenly has an extra line
- Alerting (the whole point) doesn't work with comma separated values... so now alerting is broken even for the standard osupdates...

SourceDoctor:
yeah updateing RRD is not such easy so do it in a separate RRD

you have to update osupdate Poller in LibreNMS also, so i splits  by comma

wimg:
> yeah updateing RRD is not such easy so do it in a separate RRD
> 
> you have to update osupdate Poller in LibreNMS also, so i splits by comma

I did that, but the RRD is a problem of course. Will look into that later this week.

mpikzink:
Very cool feature!

Some Idea for APT
```
apt-get --just-print upgrade |awk '/standard security updates/ {print $1}'
```

SourceDoctor:
@mpikzink 
seems to be usefull
@wimg 
could you implemnt this also?

so apt and yum are supporting this feature ...

wimg:
> @wimg
> could you implemnt this also?

I will, once I get to it. Currently swamped with other things, so it will take up to a few weeks.

mpikzink:
Any news here? I am already waiting eagerly for the PR :-)

wimg:
After using it for a while, we disabled it. It's often causing timeouts on SNMP, so I wouldn't recommend anything like this. Sadly it seems there won't be a PR coming anymore from us, since it's causing too many problems.
It might be best to look for an alternative solution.

VVelox:
> After using it for a while, we disabled it. It's often causing timeouts on SNMP, so I wouldn't recommend anything like this. Sadly it seems there won't be a PR coming anymore from us, since it's causing too many problems. It might be best to look for an alternative solution.

There is a easy fix for this. Run it via cron every 5 minutes every so often and dump the output to a file. Then just have snmpd cat that file.

Sneck, Opensearch, and several of mine require something similar to this as there is a the likely hood of them completing in a timely manner is very unlikely.

Issue #323, opened 2020-07-24
better nvme support


evanrich:
Latest commit for SMART says add support for nvme, seeing how it should pull power cycles, but it doesn't seem to work on optane drives, despite having the same wording:

```
smartctl -A /dev/nvme0
smartctl 7.0 2018-12-30 r4883 [FreeBSD 11.3-RELEASE-p9 amd64] (local build)
Copyright (C) 2002-18, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF SMART DATA SECTION ===
SMART/Health Information (NVMe Log 0x02)
Critical Warning:                   0x00
Temperature:                        41 Celsius
Available Spare:                    100%
Available Spare Threshold:          0%
Percentage Used:                    0%
Data Units Read:                    12,602,051 [6.45 TB]
Data Units Written:                 459,682,008 [235 TB]
Host Read Commands:                 125,287,911
Host Write Commands:                6,911,909,418
Controller Busy Time:               2,268
Power Cycles:                       23
Power On Hours:                     15,577
Unsafe Shutdowns:                   2
Media and Data Integrity Errors:    0
Error Information Log Entries:      0
```

when you run it though, it doesn't output "Power Cycles" as 23:

```
nvme0,null,null,null,null,null,null,null,null,null,41,null,null,null,null,null,null,0,0,0,0,0,0,0,0,15577
```

only the temperature (41) and power on hours (15577) get recorded

is it possible to also add available spare/percentage used as well?

evanrich:
here's the full smartctl output if it helps:

```
smartctl -a /dev/nvme0
smartctl 7.0 2018-12-30 r4883 [FreeBSD 11.3-RELEASE-p9 amd64] (local build)
Copyright (C) 2002-18, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF INFORMATION SECTION ===
Model Number:                       INTEL SSDPED1D280GA
Serial Number:                      PHMB7392002X280CGN
Firmware Version:                   E2010435
PCI Vendor/Subsystem ID:            0x8086
IEEE OUI Identifier:                0x5cd2e4
Controller ID:                      0
Number of Namespaces:               1
Namespace 1 Size/Capacity:          280,065,171,456 [280 GB]
Namespace 1 Formatted LBA Size:     512
Local Time is:                      Thu Jul 23 23:54:24 2020 PDT
Firmware Updates (0x02):            1 Slot
[...]

PR #383, opened 2021-10-26
- added equivalent wear level for nvme ssd

zamolxe:
- remove touched cache file to avoid no data if config is guessed
- take only 1st raw response to avoid taking strings instead of int (eg. adacom devices like supermicro sata dom moule)
- added wear level for nvme ssd's (addresses some of #323)
- use case insensitive search for disk serial numbers


Jellyfrog:
Please fix script indentation

PR #418, opened 2022-08-05
add script to monitor redis application through the check_mk agent

earendilfr:
It's the same script than `snmp/redis.py` but I just add the output of the line for application name:
```php
output = {
    "version": version,
    "error": error,
    "errorString": error_string,
    "data": redis_data,
}
print("<<<redis>>>")
print(json.dumps(output))
```


bestlong:
LGTM

PR #432, opened 2022-11-01
Add wireguard script

bnerickson:
This script is the first part in a couple of PRs to add a Wireguard application to LibreNMS.  This script runs a "wg" command, then parses all of the clients, traffic data, and last handshake data for each Wireguard interface.  Ideally, this will be displayed to the user on both the Wireguard server's interface page as-well-as the application-specific page.  For now, here's how the latter looks:

wireguard

I went back-and-forth on whether to just use the public key as the identifier for clients and add the "friendly name" as an optional parameter.  However, in the interest of security (sending public keys unencrypted over SNMPv2), I dropped that idea and am forcing users to define the mappings themselves.

Example script output:
```
{"errorString": "", "error": 0, "version": 1, "data": {"wg0": {"client1.domain.net": {"minutes_since_last_handshake": 1, "bytes_rcvd": 7060288, "bytes_sent": 119482568}, "client2.domain.net": {"minutes_since_last_handshake": null, "bytes_rcvd": 0, "bytes_sent": 0}, "client3": {"minutes_since_last_handshake": null, "bytes_rcvd": 0, "bytes_sent": 0}, "computer1": {"minutes_since_last_handshake": null, "bytes_rcvd": 0, "bytes_sent": 0}, "server": {"minutes_since_last_handshake": null, "bytes_rcvd": 0, "bytes_sent": 0}}}}
```

### DIFF
--- a/snmp/wireguard.py
+++ b/snmp/wireguard.py
@@ -43,8 +43,6 @@ import sys
 from datetime import datetime
 from itertools import chain
 
-from pathvalidate import sanitize_filename
-
 CONFIG_FILE = "/etc/snmp/wireguard.json"
 WG_ARGS = ["show", "all", "dump"]
 WG_CMD = "/usr/bin/wg"
@@ -114,7 +112,7 @@ def config_file_validator(interface_clients_dict):
     for interface, public_key_to_arbitrary_name in interface_clients_dict.items():
         rev_dict = {}
         for public_key, arbitrary_name in public_key_to_arbitrary_name.items():
-            rev_dict.setdefault(sanitize_filename(str(arbitrary_name)), set()).add(
+            rev_dict.setdefault(str(arbitrary_name), set()).add(
                 public_key
             )
 
--- a/snmp/wireguard.py
+++ b/snmp/wireguard.py
@@ -27,9 +27,9 @@
 #             "public_key_to_arbitrary_name": {
 #                 "wg0": {
 #                     "z1iSIymFEFi/PS8rR19AFBle7O4tWowMWuFzHO7oRlE=": "client1",
-#                     "XqWJRE21Fw1ke47mH1yPg/lyWqCCfjkIXiS6JobuhTI=": "server.domain.com",
+#                     "XqWJRE21Fw1ke47mH1yPg/lyWqCCfjkIXiS6JobuhTI=": "server.domain.com"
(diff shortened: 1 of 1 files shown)


Npeca75:
but why python ???

think about OpenWRT devices
they are limited in flash size
LibreNMS have already support for OpenWRT, so it would be shame to left out WG from OpenWRT because interpreter could not fit in limited flash

is there any chance to rewrite this in ASH ?
yes, ASH is default in OpenWRT

bnerickson:
> but why python ???
> 
> think about OpenWRT devices they are limited in flash size LibreNMS have already support for OpenWRT, so it would be shame to left out WG from OpenWRT because interpreter could not fit in limited flash
> 
> is there any chance to rewrite this in ASH ? yes, ASH is default in OpenWRT, and it will run nicely on full Linux bash environment

No, this will not be rewritten in ASH. Anyone interested in writing a version that can run on OpenWRT is welcome to do so using my code as a baseline, however.

To be clear, this is aimed at supporting a Linux server running WireGuard that runs Python3.

Jellyfrog (approved):

coloripari:
Hi,
thanks for this code, I'm testing on my VM but its seams dosent work on Rocky Linux 8.7.
My wireguard wgX.conf is like this:

`[Interface]
Address = x.x.x.x
ListenPort = NNNNN
PrivateKey = server_private_key_xxxxxxxxxxx

[Peer]
# Name = PerrName
PublicKey = peer_public_key_zzzzzzzzzzzz
AllowedIPs = y.y.y.y
Endpoint = ServerIP:port
PersistentKeepalive = 25
`
So in the json file I've used "peer_public_key_zzzzzzzzzzzz" and ="PeerName" but the output is always:

{"errorString": "Command Output Parsing Error: 'list index out of range'", "error": 1, "version": 1, "data": {}}

simion-chis:
@hs123org Thank you for this solution. Worked for me on Debian 12. Now I can see the graphs
```shell
sudo -u Debian-snmp /etc/snmp/wireguard.py

#output 
{"errorString": "", "error": 0, "version": 1, "data": {"wg0": {"peerName": {"minutes_since_last_handshake": 1391, "bytes_rcvd": 169184, "bytes_sent": 410684}}}}


```

PR #435, opened 2022-11-08
snmp: script: add snmpd extend for counting security updates

netravnen:
alongside the total osupdate counting


SourceDoctor:
nice idea,
but i think it would be more comfortable to enhance os_updates script with this data

adamus1red:
> nice idea, but i think it would be more comfortable to enhance os_updates script with this data

similar to #373?

PR #447, opened 2022-12-09
Cleaning up certificate.py code and adding cert_location

bnerickson:
This looks like a large change, but it's really just code cleanup and the addition of a new feature (cert_location) to support self-signed certificates.  First, the code cleanup:

I replaced the "e" variable with "err" for some pylint fixing.  The global tasks have been placed into a main function as better code practice.  Everything was run through vercel after the fact for code formatting.

Next, the cert_location feature is an optional addition to the certificate.json file that specifies the full path to the certificate for the domain/fqdn that is being validated.  This is necessary for self-signed certificates.

### DIFF
--- a/snmp/certificate.py
+++ b/snmp/certificate.py
@@ -13,8 +13,7 @@ CONFIGFILE = "/etc/snmp/certificate.json"
 # }
 
 
-def get_certificate_data(domain, port=443):
-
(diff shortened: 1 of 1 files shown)


bnerickson:
Documentation PR: https://github.com/librenms/librenms/pull/14721

murrant (approved):

PR #449, opened 2022-12-19
Address script failure with parameters

arrmo:
The current script fails to run if any parameters are set for the interface, like (for example), FwMark ... the reason is,

wg show all dump

outputs this info as well, which doesn't work with the current logic.

### DIFF
--- a/snmp/wireguard.py
+++ b/snmp/wireguard.py
@@ -234,11 +234,12 @@ def main():
         # Parse each line and import the resultant dictionary into output_data.  We update the
         # interface key with new clients as they are found and instantiate new interface keys as
         # they are found.
-        for intf, intf_data in output_parser(line, interface_clients_dict).items():
-            if intf not in output_data["data"]:
-                output_data["data"][intf] = {}
-            for client, client_data in intf_data.items():
-                output_data["data"][intf][client] = client_data
+        if len(line.strip().split()) == 9:
+            for intf, intf_data in output_parser(line, interface_clients_dict).items():
+                if intf not in output_data["data"]:
+                    output_data["data"][intf] = {}
+                for client, client_data in intf_data.items():
+                    output_data["data"][intf][client] = client_data
 
     print(json.dumps(output_data))
 


bnerickson:
Thanks, @arrmo 

Can you share an example of the sanitized `wg show all dump` output as-is?

arrmo:
Sure, NP! But ... is there an easy way to sanitize (vs. manual)?

bnerickson:
@arrmo I only want to see one line of output with the additional parameter displayed, so just one line sanitized (manually) will do.  Any line except the very first line that is.

arrmo:
Sure, NP! Here you go - make sense?
```
wg0     ---removed---    ---removed---    ---wg0Port---   off
wg1     ---removed---    ---removed---    ---wg2Port---   0xbb74
```
`0xbb74` is the fwmark. Nothing else on these two lines (in the output).

Thanks!

bnerickson:
Thanks @arrmo .  Okay, that makes sense.  Your change definitely works, but the bug is actually because I filter out the first line which is the interface-specific dump (see https://github.com/librenms/librenms-agent/blob/master/snmp/wireguard.py#L231).  Since I only have one wireguard interface, I didn't catch what would happen with multiple interfaces.

Instead of dumping with "all" or checking that the number of columns equals "9", I'll update the script to grab the list of wireguard interfaces with `wg show interfaces`, then execute a `wg show <wg_interface> dump` for each interface discovered.  This way if the number of columns for the client endpoint ever changes (exceeds or is less than 9), the script won't break (hopefully).  I can get around to doing this in a couple weeks after the holidays.

arrmo:
Works for me, thanks! Your solution makes sense to me - and yes, that `wg1` line is about half way through the output.

Have a great break!

Jellyfrog:
> I can get around to doing this in a couple weeks after the holidays.

So should this be closed then?

bnerickson:
@Jellyfrog Yes, this should be closed.

@arrmo If you want to test out my fixes, they are here: https://github.com/librenms/librenms-agent/pull/452

arrmo:
> @arrmo If you want to test out my fixes, they are here: #452

Works great, thanks!

PR #450, opened 2023-01-02
Improve docker stats

hvanoch:
https://github.com/librenms/librenms/pull/14767


snmp/docker-stats.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/python3
```
Jellyfrog, line 1:
```suggestion
#!/usr/bin/env python3
```

snmp/docker-stats.py
```diff
@@ -0,0 +1,105 @@
...
+import subprocess
+
+from dateutil import parser
+
+VERSION = 2
+ONLY_RUNNING_CONTAINERS = True
+
+
+def run(cmd):
+    res = subprocess.check_output(cmd, shell=True, stderr=subprocess.STDOUT)
```
Jellyfrog, line 13: `shell=True` looks unnecessary?

hvanoch: If I don't include shell=True I get the error
```

Traceback (most recent call last):
  File "/etc/snmp/docker-stats.py", line 105, in <module>
    dump()
  File "/etc/snmp/docker-stats.py", line 40, in dump
    stats_containers = getStats()
  File "/etc/snmp/docker-stats.py", line 28, in getStats
    raw = run(command)
  File "/etc/snmp/docker-stats.py", line 13, in run
    res = subprocess.check_output(cmd,  stderr=subprocess.STDOUT)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1823, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'docker stats --no-stream --no-trunc --format  "{{ json . }}"'

```

Jellyfrog: because you need to give it a list then, rather than a string. https://docs.python.org/3/library/subprocess.html

hvanoch: Ok my bad. I have updated it now.

PR #466, opened 2023-05-16
Fix osupdate script broken by #444

jepke:
#444  broke osupdate script through snmp extend

### DIFF
--- a/snmp/osupdate
+++ b/snmp/osupdate
@@ -35,7 +35,6 @@ CMD_APK=' version'
 ################################################################
 # Don't change anything unless you know what are you doing     #
 ################################################################
-echo '<<<app-os-updates>>>'
 if command -v zypper &>/dev/null ; then
     # OpenSUSE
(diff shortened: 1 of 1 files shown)


JamesMenetrey:
Hello,

While I'm unsure this is the right approach to fix this issue, I confirm this patch solves an issue where LibreNMS cannot pull the number of updates properly.

@VVelox This would be cool if you could look into it, as in the current state, the osupdate does not work.

Cheers!

VVelox:
@JamesMenetrey Sure! Looking.

VVelox:
Confirmed.  This only works with the older extend with out that line. The LibreNMS commit in relationship to this only passed as there are no relevant tests to make sure it did not fail.

VVelox:
Added a line to uncomment to use it as a agent. https://github.com/librenms/librenms-agent/pull/471

JamesMenetrey:
Many thanks for your quick feedback!

PR #469, opened 2023-05-28
Allow basic authentication for apache-stats.py

westerterp:
This MR allows Basic authenticaton for the Apache Statistics.


SourceDoctor:
nice,
but why not with an optional config file, so code doesn't need to be touched?
This way is uncomfortable for automatic Agent script updates

PR #474, opened 2023-06-20
add smart-v1, a new JSON based smart poller

VVelox:
Now also grabs....

- general health status
- FW version
- selftest log
- make
- model
- disk + serial

### DIFF
--- a/snmp/smart-v1
+++ b/snmp/smart-v1
@@ -1,0 +1,537 @@
+#!/usr/bin/env perl
+#Copyright (c) 2023, Zane C. Bowers-Hadley
+#All rights reserved.
+#
+#Redistribution and use in source and binary forms, with or without modification,
+#are permitted provided that the following conditions are met:
+#
+#   * Redistributions of source code must retain the above copyright notice,
+#    this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+#THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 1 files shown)


bnerickson:
Hello @VVelox ,

Is there any particular reason why you used the Gzip::Faster perl plugin as opposed to the PerlIO::gzip perl plugin?  FWIW I'm not familiar with either, so it may be that the latter isn't a drop-in replacement anyway.

I ask because RPM-based distros appear to lack broad package support for [Gzip::Fater](https://pkgs.org/download/perl-Gzip-Faster) while supporting [PerlIO::gzip](https://pkgs.org/download/perl-PerlIO-gzip) across a wider spectrum of distros.  I'd personally rather install the modules via yum/dnf instead of cpan.

VVelox:
> Hello @VVelox ,
> 
> Is there any particular reason why you used the Gzip::Faster perl plugin as opposed to the PerlIO::gzip perl plugin? FWIW I'm not familiar with either, so it may be that the latter isn't a drop-in replacement anyway.
> 
> I ask because RPM-based distros appear to lack broad package support for [Gzip::Faster](https://pkgs.org/download/perl-Gzip-Faster) while supporting [PerlIO::gzip](https://pkgs.org/download/perl-PerlIO-gzip) across a wider spectrum of distros. I'd personally rather install the modules via yum/dnf instead of cpan.

Because PerlIO is designed around file IO and not string handling.

Yeah, this is one area where how terrible RPM and DPKG based systems really are. They are a major PITA to add any new packages for given how antiquated the build system is.

VVelox:
Yeah, if you have a suggestion on one that handles strings in a similar manner and is available on the various BSDs and Linux distros, let me know. :)

I do know Gzip::Faster is available for FreeBSD via p5-Gzip-Faster and will build nicely on every Linux distro I've used it on.

VVelox:
> IO::Compress::Gzip

Thanks! Used in https://github.com/librenms/librenms-agent/pull/476 .

PR #494, opened 2023-10-21
Quick and dirty fix for snmp extend to respond correctly.

calvinthefreak:
- Added "grep -v" to the output command.
 - Added a note for Users that don't understand why LibreNMS can't get the values when snmp user is not in docker group.


VVelox:
thanks!

PR #497, opened 2023-11-21
mysql - ensure returned values do not exceed max of RRD file

peejaychilds:
Should resolve https://github.com/librenms/librenms/issues/13169


VVelox:
thanks!

PR #516, opened 2024-04-26
HTTPS with CA validation and HTTP Basic Authentication implemented

Marco-Valle:
HTTPS with CA validation and HTTP Basic Authentication with token in file implemented.

HTTPS can be enabled with the option -S

The path of the CA certificate to be used for validation can be specified with the option -c

The path of the base64 encoded username:password file can be specified with the option -a


Marco-Valle:
Close cause wrong author and issues with CLAassistant.

PR #563, opened 2025-02-17
Add hddtemp2 script as modern replacement for hddtemp

jasoncheng7115:
This commit introduces a new script 'hddtemp2' as an alternative solution for  LibreNMS disk temperature monitoring on systems where the hddtemp package is no  longer available in the default repositories.

Key features:
- Uses smartctl for SATA/SAS disk temperature monitoring
- Adds native NVMe disk temperature support 
- Works on modern Ubuntu/Debian systems without requiring hddtemp
- Maintains original LibreNMS agent output format
- Supports both traditional hard drives and NVMe devices
- Includes robust error handling and device validation

The script is designed to be a drop-in replacement for hddtemp functionality  in LibreNMS agent, ensuring seamless integration with existing monitoring  setups while providing broader hardware support.

Testing:
- Verified on Ubuntu/Debian systems where hddtemp is unavailable
- Tested with various SATA, SAS, and NVMe drives
- Confirmed compatibility with LibreNMS agent format

### DIFF
--- a/agent-local/hddtemp2
+++ b/agent-local/hddtemp2
@@ -3,6 +3,7 @@
 # Name: hddtemp2
 # Author: Jason Cheng (www.jason.tools)
 # Date: 2025-02-17
+# Version: 1.1
 # Purpose: This script replaces the obsolete hddtemp tool for librenms agent to monitor disk temperatures
 # License: Free Software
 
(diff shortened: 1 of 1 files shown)


VVelox:
thanks!

PR #564, opened 2025-02-19
Pi-hole v6 API changes

karlshea:
This is getting my stats working with v6.

I *did not* test or fix DHCP since I don't use it, and if you have an API key their endpoint requires you to get a token, so that's not working either.

I didn't think to look at the v5 API before I upgraded my instances, but it looks like it must have been outputting query types in thousands. My numbers all jumped from like ~55 to ~55,000 but they're all proportionally correct. And all of the rest of the graphs seem to be the same.

### DIFF
--- a/snmp/pi-hole
+++ b/snmp/pi-hole
@@ -5,9 +5,8 @@ IFS=$'\n\t'
 CONFIGFILE='/etc/snmp/pi-hole.conf'
 
 API_AUTH_KEY=""
-API_URL="localhost/admin/api.php"
-URL_READ_ONLY="?summaryRaw&auth="
-URL_QUERY_TYPE="?getQueryTypes&auth="
+API_URL="localhost/api"
+URL_READ_ONLY="/stats/summary"
 PICONFIGFILE='/etc/pihole/setupVars.conf'
 DHCPLEASEFILE='/etc/pihole/dhcp.leases'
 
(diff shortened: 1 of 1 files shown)


SourceDoctor:
very nice work.

But would it be possible to integrate a version switch, so pihole5 is also supported?

Especially for the migration phase?

karlshea:
I'm not really sure what the right way to do that would be while also allowing individual config overrides. You'd want a VERSION variable, but then all of the other variables would have to change based on it. However you'd want to read in VERSION from the config, so that would need to be loaded first, which means the rest couldn't be overridden.

The only alternative would be separate variables for each version.

SourceDoctor:
@karlshea 

i tested it on my pihole6, but it seems not to work

Looking into API Documentation says, there is a request with password to /api/auth needed to get a sid.
With this sid real request can be started.

Did i miss something, or is something missing in Code change?

greetings

karlshea:
Yes, I warned about this:

> if you have an API key their endpoint requires you to get a token, so that's not working either.

I was more looking for some testing and review, I'm not sure it was quite ready to go in yet.

SourceDoctor:
It needs to be overlooked completely.

Maybe a full rewrite would be the best and fastest way, to meet the new authentication style, and also fullfill the actual librenms application data communication structure.

Best and easiest way is to be done with python3.
Are you familiar with it?
If not, i will do it, because there are some Code changes in LibreNMS itself also.

Nevertheless, your code change is a good entry point to point to the done changes in api response.

karlshea:
Oh, I have a PR open: #566 

I don't know Python well enough to rewrite it.

PR #569, opened 2025-05-09
Fix typo in comment

sshockley:
This fixes a minor typo in osupdates-unpriv-generate.sh.

### DIFF
--- a/snmp/unpriv/osupdates/osupdates-unpriv-gather.sh
+++ b/snmp/unpriv/osupdates/osupdates-unpriv-gather.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-SNMP_PERSISTENT_DIR="$(net-snmp-config --persistent-directory)"
+SNMP_PERSISTENT_DIR="/var/lib/net-snmp"
 UNPRIV_SHARED_FILE="$SNMP_PERSISTENT_DIR/osupdates/stats.txt"
 
 if [ -f "$UNPRIV_SHARED_FILE" ]; then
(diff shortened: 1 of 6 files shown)


snmp/unpriv/osupdates/osupdates-unpriv-gather.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-SNMP_PERSISTENT_DIR="$(net-snmp-config --persistent-directory)"
+SNMP_PERSISTENT_DIR="/var/lib/net-snmp"
```
dlangille, line 3: This seems unintentional

snmp/unpriv/osupdates/osupdates-unpriv-generate.sh
```diff
@@ -29,7 +29,7 @@ BIN_PKG='/usr/sbin/pkg'
 CMD_PKG=' audit -q -F'
 BIN_APK='/sbin/apk'
 CMD_APK=' version'
-SNMP_PERSISTENT_DIR="$(net-snmp-config --persistent-directory)"
+SNMP_PERSISTENT_DIR="/var/lib/net-snmp"
```
dlangille, line 32: Oh, perhaps not.  but why?

dlangille (commented):
Wondering why a move from a function call to a hardcoded value which may vary under other operating systems.

snmp/unpriv/osupdates/osupdates-unpriv-generate.sh
dlangille: So Fedora doesn't have it, so the rest of the OS must accommodate it?

Why is Fedora the chosen correct install path?

dlangille (commented):
I don't see why we're allowing a change because a given OS doesn't like what was there.

sshockley:
Oh, it grabbed everything I did after the typo fix too, even though I created the PR before I made those changes.  I'll close this and try again, sorry for the noise.

PR #582, opened 2025-08-18
read battery.temperature or ups.temperature

osiktech:
I have two UPSs from APC. One reports battery.temperature the other ups.temperature. This small fix should fix to get a temperature reading.


VVelox (approved):

VVelox:
@osiktech derp! missed that =^.^=

PR #595, opened 2026-02-22
OpenWrt agent: add temperature sensors and harden wireless SNMP setup - ignore that PR

perceival:
This change is adding temperature sensors and improving (creating automation) in SNMP setup for WiFi sensors. Dependent on https://github.com/librenms/librenms/pull/19092


perceival:
Quick refresh summary:

- Removed OpenWrt standalone README from agent repo (moved guidance into LibreNMS OpenWrt docs).
- Made `setup-snmpd.sh` idempotent so generated wireless extends are replaced, not appended (prevents duplicate stale entries).
- Fixed generator format for OpenWrt UCI compatibility and updated `wlClients.sh` counting/normalization (including aggregate dedupe behavior for multi-link scenarios).

Validated on-device: `nsExtendCommand` now shows clean current interface entries only, and client extends return valid values without script errors.
@VVelox could you please re-review the PR?

PR #608, opened 2026-03-10
add new app: systemd-timesyncd

MarlinMr:
New extend file to put in `/etc/snmp/`.

See https://github.com/librenms/librenms/pull/19198 for more details


VVelox:
thanks!

PR #611, opened 2026-03-11
remove ntp-client and ntpq related comments

MarlinMr:
There were left comments relating to functionality in `ntp-client`. Had forgotten to remove those in https://github.com/librenms/librenms-agent/pull/608


VVelox:
derp, missed that, thanks!

PR #612, opened 2026-03-11
One ntp to rule

MarlinMr:
- [x] agree on variables to send back
- [x] extend to functionality from ntp-server.py

### DIFF
--- a/snmp/ntp
+++ b/snmp/ntp
@@ -1,0 +1,71 @@
+#!/bin/sh
+# Please make sure the paths below are correct.
+# Alternatively you can put them in $0.conf, meaning if you've named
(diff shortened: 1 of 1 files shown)


MarlinMr:
See https://github.com/librenms/librenms/pull/19291

VVelox (changes requested):
Needs to be portable. Use a flag for which should be used.

snmp/ntp-client
VVelox, line 1: Also we generally leave older versions of the extends laying around if they work in case any one has anything that depends on those as part of a ansible/rex job or the like.

VVelox (changes requested):
Generally best to leave the old ones in place still if they work.

PR #613, opened 2026-03-13
OpenWrt agent: add temperature sensors and harden wireless SNMP setup

perceival:
This change is adding temperature sensors and improving (creating automation) in SNMP setup for WiFi sensors. Dependent on https://github.com/librenms/librenms/pull/19347 (recreated after botched previous PR - https://github.com/librenms/librenms-agent/pull/595).

Notes for reviewers:

[wlInterfaces.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Replaces static interface mapping with live discovery from hostapd/iwinfo.
Filters inactive/placeholder VAPs and prefers active hostapd-managed interfaces.
Produces stable labels and handles duplicate SSIDs by appending band suffixes (24/5/6).
Keeps fallback logic for systems without full hostapd metadata.
[wlClients.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Moves from simple station-dump counting to MAC-based counting with dedupe.
Supports both per-interface and aggregate modes.
Uses hostapd get_clients when available, with iwinfo fallback.
Deduplicates MLO/MLD-related addresses in aggregate mode to avoid overcounting.
[wlRate.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [wlNoiseFloor.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [wlSNR.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [wlFrequency.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Adds fallback parsing paths for client-mode interfaces where assoclist can be empty.
Normalizes output for SNMP extend compatibility (first line numeric value, optional second context line).
[setup-snmpd.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [snmpd-config-generator.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Introduces install/apply workflow for OpenWrt helper scripts.
Ensures managed LibreNMS blocks are replaced cleanly instead of duplicated.
Supports non-interactive apply (-y) and optional no-restart (--no-restart).
[lm-sensors-pass.sh](vscode-file://vscode-app/usr/share/code-insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Adds pass script implementing LM-SENSORS-MIB subtree behavior for thermal zones.
Handles both GET and GETNEXT traversal semantics for SNMP requests.
Scope check

PR remains constrained to OpenWrt integration paths with one upstream-based commit and the expected 12-file diff.

@VVelox I have recreated the PR from the scratch including your requested changes. Kindly please review.


perceival:
@VVelox could you please take a look...?

perceival:
Updated the net-snmp OpenWrt integration: the distro extend now uses /etc/os-release for improved detection, and the unused snmpd-base-config file has been removed.
@VVelox kindly please review.

perceival:
Opened #618 to give this broader visibility and track the feature request separately from the implementation. The PR remains ready for review.

perceival:
@VVelox could  you...?

sshockley:
For what it's worth I'm testing this now and it's working well.

perceival:
Friendly bump @VVelox — @sshockley's report above ("testing this now and it's working well") confirms the agent works in a second OpenWrt environment beyond my own. PR is mergeable, no conflicts, CLA signed. Could you take a look when you have a minute? Happy to address any review feedback.

snmp/Openwrt/LICENSE
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
laf: Remove this please

snmp/Openwrt/snmpd-config-generator.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/sh
```
laf: This whole file needs removing.

laf (changes requested):
I honestly think this needs to be done using an LibreNMS MIB and snmpd pass_persist or possibly AgentX.

perceival:
Thanks for the review, @laf!

- `snmp/Openwrt/LICENSE` — dropped (stray, agreed).
- `snmp/Openwrt/snmpd-config-generator.sh` — this is the helper that generates the snmpd `extend` config so users don't hand-edit `snmpd.conf`, so its fate is really part of the config approach. If we stay on `extend` I'm happy to drop the generator and document the snmpd lines for users to add manually instead; if we move to `pass_persist`/MIB it's moot anyway. Which leads to the main question:

On **MIB + `pass_persist`/AgentX vs the current `extend` approach** — I'd value your steer before building, since it cascades into the server-side PR #19347 too:

- `extend` is what most existing LibreNMS agent scripts use, it drops in with no extra moving parts for users, and it's been validated in two independent OpenWrt deployments (mine + @sshockley's "working well" report above).
- That said, I'm genuinely happy to implement a proper MIB + `pass_persist` (or AgentX) if that's the bar for inclusion — it's just a substantial rework across both the agent and the LibreNMS-side discovery, so I'd rather confirm the direction than build it twice.

So: is `pass_persist`/MIB a hard requirement, or would a cleaned-up `extend` implementation (incl. the `snmp_*`→`SnmpQuery` conversion you flagged on #19347) be acceptable? Either way I'll take it there. Thanks!

snmp/Openwrt/LICENSE
perceival: Done — removed in aa7ab05e.

snmp/Openwrt/snmpd-config-generator.sh
perceival: Removed `snmpd-config-generator.sh` (and `setup-snmpd.sh`). The generator now lives as an adaptable sample in the new `snmp/Openwrt/README.md`, with a safe re-apply using `AUTOGEN` markers — documented but no longer shipped as a managed file. (9275fabf)

Torstein-Eide:
@perceival 

I think it should be something like this:

you can take a look at 

[snmp/xcp-ng-vminfo](https://github.com/librenms/librenms-agent/blob/master/snmp/xcp-ng-vminfo)
or 
[SNMP: MDADM agent and mib](https://github.com/Torstein-Eide/librenms-agent/tree/multiple-Device-md-agent/snmp/mdadm)

it can be also good to look at [mibs.observium.org](https://mibs.observium.org/all) if there is othere wireless metric to be included

```asn
OPENWRT-MIB DEFINITIONS ::= BEGIN

IMPORTS
    MODULE-IDENTITY, OBJECT-TYPE,
    Integer32, Unsigned32, enterprises
        FROM SNMPv2-SMI
    DisplayString
        FROM SNMPv2-TC
    MODULE-COMPLIANCE, OBJECT-GROUP
        FROM SNMPv2-CONF
    InterfaceIndex
        FROM IF-MIB;

openwrtMIB MODULE-IDENTITY
    LAST-UPDATED "202606250000Z"
    ORGANIZATION "Unassigned"
    CONTACT-INFO
        "OpenWrt SNMP agent MIB.

         Enterprise OID: 1.3.6.1.4.1.60652.102.0 "
    DESCRIPTION
        "The OPENWRT-MIB describes OpenWrt wireless metrics exported by
         the LibreNMS OpenWrt SNMP extension scripts."
    REVISION "202606250000Z"
    DESCRIPTION "Initial version for LibreNMS OpenWrt wireless extends."
    ::= { enterprises 60652 102 }

-- --------------------------------------------------------------------------
-- Top-level structure
-- --------------------------------------------------------------------------

openwrtObjects     OBJECT IDENTIFIER ::= { openwrtMIB 1 }
openwrtConformance OBJECT IDENTIFIER ::= { openwrtMIB 2 }
openwrtCompliances OBJECT IDENTIFIER ::= { openwrtConformance 1 }
openwrtGroups      OBJECT IDENTIFIER ::= { openwrtConformance 2 }

--- Maybe othere openwrt metadata

-- --------------------------------------------------------------------------
-- Section 10: Wireless metadata and aggregate scalars
-- --------------------------------------------------------------------------

openwrtWireless OBJECT IDENTIFIER ::= { openwrtObjects 10 }

openwrtWirelessInterfaceCount OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Number of active wireless interfaces"
    ::= { openwrtWireless 1 }

openwrtWirelessClientCount OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Aggregate count of unique associated wireless clients across the
         exported interfaces.
         NET-SNMP extend source: clients-wlan"
    ::= { openwrtWireless 2 }

-- --------------------------------------------------------------------------
-- Section 10B: Wireless interface table
-- --------------------------------------------------------------------------

openwrtWirelessInterfaceTable OBJECT-TYPE
    SYNTAX      SEQUENCE OF OpenwrtWirelessInterfaceEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "OpenWrt wireless interfaces and per-interface metrics."
    ::= { openwrtWireless 3 }

openwrtWirelessInterfaceEntry OBJECT-TYPE
    SYNTAX      OpenwrtWirelessInterfaceEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "Metrics for one OpenWrt wireless interface.  Rows are indexed by
         the corresponding IF-MIB::ifIndex value."
    INDEX { openwrtWlIfaceIfIndex }
    ::= { openwrtWirelessInterfaceTable 1 }

OpenwrtWirelessInterfaceEntry ::= SEQUENCE {
    openwrtWlIfaceIfIndex    InterfaceIndex,
    openwrtWlIfaceName       DisplayString,
    openwrtWlIfaceLabel      DisplayString,
    openwrtWlIfaceClients    Unsigned32,
    openwrtWlIfaceFrequency  Unsigned32,
    openwrtWlIfaceNoiseFloor Integer32,
    openwrtWlIfaceTxRateMin  Unsigned32,
    openwrtWlIfaceTxRateAvg  Unsigned32,
    openwrtWlIfaceTxRateMax  Unsigned32,
    openwrtWlIfaceRxRateMin  Unsigned32,
    openwrtWlIfaceRxRateAvg  Unsigned32,
    openwrtWlIfaceRxRateMax  Unsigned32,
    openwrtWlIfaceSnrMin     Integer32,
    openwrtWlIfaceSnrAvg     Integer32,
    openwrtWlIfaceSnrMax     Integer32
}

openwrtWlIfaceIfIndex OBJECT-TYPE
    SYNTAX      InterfaceIndex
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "Interface index from IF-MIB::ifIndex for this wireless interface."
    ::= { openwrtWirelessInterfaceEntry 1 }

openwrtWlIfaceName OBJECT-TYPE
    SYNTAX      DisplayString
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "OpenWrt wireless interface name, for example wlan0 or phy0-ap0."
    ::= { openwrtWirelessInterfaceEntry 2 }

openwrtWlIfaceLabel OBJECT-TYPE
    SYNTAX      DisplayString
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Display label for the wireless interface, normally the SSID.  If
         multiple interfaces share an SSID, wlInterfaces.sh appends a band
         suffix such as 24, 5, or 6."
    ::= { openwrtWirelessInterfaceEntry 3 }

openwrtWlIfaceClients OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Associated client count for this wireless interface."
    ::= { openwrtWirelessInterfaceEntry 4 }

openwrtWlIfaceFrequency OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "MHz"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Current channel frequency in MHz. "
    ::= { openwrtWirelessInterfaceEntry 5 }

openwrtWlIfaceNoiseFloor OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dBm"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Noise floor for this wireless interface in dBm. "
    ::= { openwrtWirelessInterfaceEntry 6 }

openwrtWlIfaceTxRateMin OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum transmit bitrate across associated stations."
    ::= { openwrtWirelessInterfaceEntry 7 }

openwrtWlIfaceTxRateAvg OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average transmit bitrate across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 8 }

openwrtWlIfaceTxRateMax OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum transmit bitrate across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 9 }

openwrtWlIfaceRxRateMin OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum receive bitrate across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 10 }

openwrtWlIfaceRxRateAvg OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average receive bitrate across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 11 }

openwrtWlIfaceRxRateMax OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum receive bitrate across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 12 }

openwrtWlIfaceSnrMin OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum signal-to-noise ratio across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 13 }

openwrtWlIfaceSnrAvg OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average signal-to-noise ratio across associated stations. "
    ::= { openwrtWirelessInterfaceEntry 14 }

openwrtWlIfaceSnrMax OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum signal-to-noise ratio across associated stations."
    ::= { openwrtWirelessInterfaceEntry 15 }


END
```

perceival:
Thanks @Torstein-Eide — this is exactly the steer I was hoping for, and the `xcp-ng-vminfo` / `mdadm` references are very helpful as patterns to follow.

I pulled a clean baseline walk of vanilla OpenWrt (stock `snmpd`, no extends) to see what we actually have to build on, and it backs your table design well:

- **Nothing wireless exists in standard MIBs.** IEEE 802.11 (`.1.2.840.10036`) is a single stub `RTID` line; there's no client count, frequency, noise floor, tx/rx rate or SNR anywhere. So every column in your `openwrtWirelessInterfaceTable` is genuinely new data — none of it duplicates an existing MIB.
- **The VAPs are already in IF-MIB with stable ifIndexes**, e.g. `34=wl0-ap0`, `35=wl1-ap0`, `36=wl1-ap1`, `37=wl1-ap2`, `41=wl1-ap1.sta1`. That makes indexing the table by `ifIndex` a clean join onto interface rows LibreNMS already discovers — nice call.
- **Temperature** is separate: LM-SENSORS-MIB (`.2021.13.16`) is empty on the baseline, so that rides the `lm-sensors-pass.sh` → LM-SENSORS-MIB emulation path (which the server PR #19347 already keys on and is working in two deployments). I'd keep that out of OPENWRT-MIB and leave it as-is — so this rework is wireless-only.

One design point so the join is solid: the table index needs to be the device's **real kernel `ifIndex`** (`/sys/class/net/<iface>/ifindex`), not a counter the script invents — otherwise metrics can attach to the wrong interface row. The current `wlInterfaces.sh` was written for the flat `extend` model, so I'll adapt it to emit the kernel ifIndex. Worth noting one VAP in my walk is a client/STA-mode interface (`wl1-ap1.sta1`); I'll make sure those round-trip too. Does that match how you indexed the `mdadm` table?

@VVelox @laf — before I commit to the rework (it spans both this PR and server-side #19347), could you confirm the direction? Specifically:

1. **Is the MIB + `pass_persist`/AgentX approach the bar for inclusion**, replacing the current `extend` implementation? Happy to build it — just want to confirm before doing it twice.
2. **OID sub-arc:** the draft sits under `enterprises 60652`, which is LibreNMS's registered PEN — good. Who allocates the sub-arc beneath it (Torstein's draft uses `.102`), and is there an existing registry of LibreNMS sub-arcs I should slot into so it doesn't collide with `LIBRENMS-NOTIFICATIONS-MIB`? I'll also set the MODULE-IDENTITY `ORGANIZATION`/`CONTACT-INFO` to LibreNMS rather than the `"Unassigned"` placeholder.

Once the direction and the sub-arc are settled I'll do the OPENWRT-MIB + `pass_persist` rework here and the matching table-walk discovery in #19347 in one consistent pass.

VVelox:
@perceival Yeah. Sounds good to me. Especially if it manages to avoid the mess of adding lots of stuff to the config, which is something I really hate about the current implementation for OpenWRT radio stuff... have to add a entry for every item instead one that will take care of it all.

Torstein-Eide:
> ... Does that match how you indexed the `mdadm` table?

No the `Madam` uses a 31bit hash of md UUID, since Linux reuses the same short names, even when not restarting.  

IfIndex is not unique across reboots, but other mibs uses it. There is logic in LibreNMS that groups sensors by  IfIndex.

perceival:
Progress + a couple of things to ratify before I build the agent/server sides on top.

I took @Torstein-Eide's draft and turned it into a lint-clean MIB. It's `smilint -l 6` clean (no output) and net-snmp loads the full tree. A few decisions I made, flagged so they can be confirmed rather than silently baked in:

- **Arc: `{ librenms 102 }`** (i.e. `.1.3.6.1.4.1.60652.102`). This follows the existing LibreNMS agent-MIB convention — `XCP-NG-VMINFO-MIB` sits at `{ librenms 100 }` and `{ librenms 1 }` is alerting, so OS/agent MIBs live in the 100-block. I kept Torstein's `102` (101 also looks free). **@VVelox / @laf — could you rubber-stamp the sub-arc number** so it's official before it becomes the contract both PRs depend on?
- **Node descriptor `librenms` (lower-case)** to match `LIBRENMS-NOTIFICATIONS-MIB` and SMI naming rules (`XCP-NG-VMINFO-MIB` spells it `LibreNMS`, which trips smilint — avoided here).
- **Index = IF-MIB `ifIndex`** per Torstein's guidance, so the table joins onto the standard ifTable/ifXTable rows for the same VAP.
- **Rates in Mbit/s, not bit/s.** `Unsigned32` in bit/s overflows on Wi-Fi 6E/7 (a BE9300 negotiates ~9.3 Gbit/s); Mbit/s keeps headroom. (Server-side `WirelessRateDiscovery` works in bit/s, so it'll scale on the LibreNMS end.)

**One shape question** before I wire it up: my baseline walk includes a client/STA-mode interface (`wl1-ap1.sta1`). This table is AP-centric — `openwrtWlIfaceClients` (and arguably the per-station rate/SNR aggregates) don't really mean anything for a station. Should client-mode interfaces be **included** in the table (with N/A-ish values) or **omitted** entirely? I lean toward omitting STA interfaces from this table; happy to do either. _(See follow-up below — on reflection I'd now include them; the MIB text below reflects that, with `openwrtWlIfaceClients` documented as the STA peer count 0/1.)_

Temperature stays out of this MIB (rides LM-SENSORS-MIB emulation as before).

Full MIB below for review:

```asn1
-- **********************************************************
-- OPENWRT-MIB
--
-- Custom MIB for OpenWrt wireless metrics exported by the
-- LibreNMS OpenWrt SNMP agent (pass_persist handler).
--
-- Enterprise Number: 60652 (LibreNMS, IANA PEN)
-- OID Base: .1.3.6.1.4.1.60652
--
-- NOTE: the sub-arc { LibreNMS 102 } is PROVISIONAL, following the
-- LibreNMS agent-MIB convention where OS/agent MIBs occupy the 100-block
-- (XCP-NG-VMINFO-MIB uses { LibreNMS 100 }; { LibreNMS 1 } is alerting).
-- Pending final assignment by a LibreNMS maintainer.
--
-- Temperature/thermal data is intentionally NOT defined here; it is
-- exported separately via LM-SENSORS-MIB emulation.
-- **********************************************************

OPENWRT-MIB DEFINITIONS ::= BEGIN

IMPORTS
    MODULE-IDENTITY, OBJECT-TYPE,
    Integer32, Unsigned32, enterprises
        FROM SNMPv2-SMI
    DisplayString
        FROM SNMPv2-TC
    MODULE-COMPLIANCE, OBJECT-GROUP
        FROM SNMPv2-CONF
    InterfaceIndex
        FROM IF-MIB;

openwrtMIB MODULE-IDENTITY
    LAST-UPDATED "202606250000Z"
    ORGANIZATION "LibreNMS"
    CONTACT-INFO
        "       LibreNMS Community
                https://www.librenms.org
                https://github.com/librenms/librenms
        "
    DESCRIPTION
        "MIB module describing OpenWrt wireless metrics exported by the
         LibreNMS OpenWrt SNMP agent. The agent discovers radios/VAPs live
         (hostapd/iwinfo) at request time and serves this subtree from a
         single net-snmp pass_persist handler, so no per-interface or
         per-metric snmpd configuration is required."
    REVISION "202606250000Z"
    DESCRIPTION
        "Initial version: wireless interface table and aggregate scalars."
    ::= { librenms 102 }

-- **********************************************************
-- Top-level enterprise OID
-- **********************************************************

-- Matches the `librenms' node descriptor used by LIBRENMS-NOTIFICATIONS-MIB
-- (lower-case per SMI naming rules; XCP-NG-VMINFO-MIB spells it `LibreNMS').
librenms OBJECT IDENTIFIER ::= { enterprises 60652 }

openwrtObjects     OBJECT IDENTIFIER ::= { openwrtMIB 1 }
openwrtConformance OBJECT IDENTIFIER ::= { openwrtMIB 2 }

-- **********************************************************
-- Wireless objects
-- **********************************************************

openwrtWireless OBJECT IDENTIFIER ::= { openwrtObjects 10 }

openwrtWirelessInterfaceCount OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Number of active wireless interfaces exported in
         openwrtWirelessInterfaceTable."
    ::= { openwrtWireless 1 }

openwrtWirelessClientCount OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Aggregate count of unique wireless clients served by this device
         across its AP/master interfaces. MLO/MLD-related addresses are
         de-duplicated so a single client associated over multiple links counts
         once. Upstream associations on client/STA-mode interfaces are excluded
         (this device's own uplink is not one of its clients)."
    ::= { openwrtWireless 2 }

-- **********************************************************
-- Wireless interface table
-- **********************************************************

openwrtWirelessInterfaceTable OBJECT-TYPE
    SYNTAX      SEQUENCE OF OpenwrtWirelessInterfaceEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "OpenWrt wireless interfaces and their per-interface metrics."
    ::= { openwrtWireless 3 }

openwrtWirelessInterfaceEntry OBJECT-TYPE
    SYNTAX      OpenwrtWirelessInterfaceEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "Metrics for one OpenWrt wireless interface. Rows are indexed by the
         corresponding IF-MIB ifIndex, so that this table joins directly onto
         the standard interface table for the same VAP."
    INDEX { openwrtWlIfaceIfIndex }
    ::= { openwrtWirelessInterfaceTable 1 }

OpenwrtWirelessInterfaceEntry ::= SEQUENCE {
    openwrtWlIfaceIfIndex    InterfaceIndex,
    openwrtWlIfaceName       DisplayString,
    openwrtWlIfaceLabel      DisplayString,
    openwrtWlIfaceClients    Unsigned32,
    openwrtWlIfaceFrequency  Unsigned32,
    openwrtWlIfaceNoiseFloor Integer32,
    openwrtWlIfaceTxRateMin  Unsigned32,
    openwrtWlIfaceTxRateAvg  Unsigned32,
    openwrtWlIfaceTxRateMax  Unsigned32,
    openwrtWlIfaceRxRateMin  Unsigned32,
    openwrtWlIfaceRxRateAvg  Unsigned32,
    openwrtWlIfaceRxRateMax  Unsigned32,
    openwrtWlIfaceSnrMin     Integer32,
    openwrtWlIfaceSnrAvg     Integer32,
    openwrtWlIfaceSnrMax     Integer32
}

openwrtWlIfaceIfIndex OBJECT-TYPE
    SYNTAX      InterfaceIndex
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
        "The kernel interface index (IF-MIB ifIndex) of this wireless
         interface, taken from /sys/class/net/<iface>/ifindex. Used as the
         table index so rows align with the standard ifTable/ifXTable."
    ::= { openwrtWirelessInterfaceEntry 1 }

openwrtWlIfaceName OBJECT-TYPE
    SYNTAX      DisplayString
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "OpenWrt wireless interface name, for example wlan0 or phy0-ap0."
    ::= { openwrtWirelessInterfaceEntry 2 }

openwrtWlIfaceLabel OBJECT-TYPE
    SYNTAX      DisplayString
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Display label for the wireless interface, normally the SSID. If
         multiple interfaces share an SSID, the agent appends a band suffix
         such as 24, 5 or 6 to keep labels unique."
    ::= { openwrtWirelessInterfaceEntry 3 }

openwrtWlIfaceClients OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Number of stations associated with this wireless interface.
         For an AP/master interface this is the count of connected clients.
         For a client/STA-mode interface (e.g. a wireless uplink or backhaul)
         this is the number of upstream associations, which is its peer count:
         1 while associated to an AP, 0 when there is no peer (not associated /
         link down) - so it doubles as a backhaul link-up indicator."
    ::= { openwrtWirelessInterfaceEntry 4 }

openwrtWlIfaceFrequency OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "MHz"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Current channel centre frequency in MHz."
    ::= { openwrtWirelessInterfaceEntry 5 }

openwrtWlIfaceNoiseFloor OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dBm"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Noise floor for this wireless interface in dBm (typically negative)."
    ::= { openwrtWirelessInterfaceEntry 6 }

openwrtWlIfaceTxRateMin OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum transmit bitrate across associated stations, in Mbit/s.
         Expressed in Mbit/s rather than bit/s so that Wi-Fi 6E/7 rates
         (multi-Gbit/s) do not overflow Unsigned32."
    ::= { openwrtWirelessInterfaceEntry 7 }

openwrtWlIfaceTxRateAvg OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average transmit bitrate across associated stations, in Mbit/s."
    ::= { openwrtWirelessInterfaceEntry 8 }

openwrtWlIfaceTxRateMax OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum transmit bitrate across associated stations, in Mbit/s."
    ::= { openwrtWirelessInterfaceEntry 9 }

openwrtWlIfaceRxRateMin OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum receive bitrate across associated stations, in Mbit/s."
    ::= { openwrtWirelessInterfaceEntry 10 }

openwrtWlIfaceRxRateAvg OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average receive bitrate across associated stations, in Mbit/s."
    ::= { openwrtWirelessInterfaceEntry 11 }

openwrtWlIfaceRxRateMax OBJECT-TYPE
    SYNTAX      Unsigned32
    UNITS       "Mbit/s"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum receive bitrate across associated stations, in Mbit/s."
    ::= { openwrtWirelessInterfaceEntry 12 }

openwrtWlIfaceSnrMin OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Minimum signal-to-noise ratio across associated stations, in dB."
    ::= { openwrtWirelessInterfaceEntry 13 }

openwrtWlIfaceSnrAvg OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Average signal-to-noise ratio across associated stations, in dB."
    ::= { openwrtWirelessInterfaceEntry 14 }

openwrtWlIfaceSnrMax OBJECT-TYPE
    SYNTAX      Integer32
    UNITS       "dB"
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "Maximum signal-to-noise ratio across associated stations, in dB."
    ::= { openwrtWirelessInterfaceEntry 15 }

-- **********************************************************
-- Conformance
-- **********************************************************

openwrtCompliances OBJECT IDENTIFIER ::= { openwrtConformance 1 }
openwrtGroups      OBJECT IDENTIFIER ::= { openwrtConformance 2 }

openwrtWirelessScalarGroup OBJECT-GROUP
    OBJECTS {
        openwrtWirelessInterfaceCount,
        openwrtWirelessClientCount
    }
    STATUS  current
    DESCRIPTION
        "Aggregate wireless scalars."
    ::= { openwrtGroups 1 }

openwrtWirelessInterfaceGroup OBJECT-GROUP
    OBJECTS {
        openwrtWlIfaceName,
        openwrtWlIfaceLabel,
        openwrtWlIfaceClients,
        openwrtWlIfaceFrequency,
        openwrtWlIfaceNoiseFloor,
        openwrtWlIfaceTxRateMin,
        openwrtWlIfaceTxRateAvg,
        openwrtWlIfaceTxRateMax,
        openwrtWlIfaceRxRateMin,
        openwrtWlIfaceRxRateAvg,
        openwrtWlIfaceRxRateMax,
        openwrtWlIfaceSnrMin,
        openwrtWlIfaceSnrAvg,
        openwrtWlIfaceSnrMax
    }
    STATUS  current
    DESCRIPTION
        "Per-interface wireless metrics."
    ::= { openwrtGroups 2 }

openwrtCompliance MODULE-COMPLIANCE
    STATUS  current
    DESCRIPTION
        "Compliance statement for OpenWrt wireless monitoring agents."
    MODULE  -- this module
    MANDATORY-GROUPS {
        openwrtWirelessScalarGroup,
        openwrtWirelessInterfaceGroup
    }
    ::= { openwrtCompliances 1 }

END
```

perceival:
Correction on my own STA-mode lean above — having thought about the device class, I'd now argue the opposite: **include STA/client-mode interfaces** in the table.

On an OpenWrt AP/router a STA interface is almost always a wireless uplink/backhaul (mesh node, repeater, wireless WAN), and that's exactly the link where rate, SNR and noise floor are the *most* worth monitoring — arguably more than on a serving AP VAP. Practically it's just another wireless interface that needs watching. The only column that's vacuous for a STA is `openwrtWlIfaceClients`, which simply reports 0; everything else applies cleanly. So no MIB change needed — I'll just not filter STA rows out.

If it's useful to tell AP vs STA (vs mesh) apart on the LibreNMS side, one option is an optional `openwrtWlIfaceMode` column (ap/sta/mesh/monitor) — IF-MIB `ifType` won't distinguish them. Happy to add it or leave it out; flagging it now since we're settling the table shape. @Torstein-Eide @VVelox — any preference?

Torstein-Eide:
1. For dialog, like Pull request. avoid using LLM directly. The maintainers are overload with low "quality" pull request, so a way to standout is not give the impression that one is talking to a LLM. 

### For SNMP mib

1. I wound at minimum add Metadata (.0) 
  a.  0.1: openwrtMIBversion
  b. 0.2: openwrtOSversion
2.  I wound add  Firewall 2.
  a. metatadata. 
  b. table of firewall rules
  c. table of hitrate
  d. table of active sesions.
3. I wound add DHCP 3
 a. metadata 
 b. table of dhcp servers
 c. table of dhcp clients 

and so on.

perceival:
ad. 1. Good point. Turning off lazy mode ;-)
Regarding MIBs:
My approach would be to have it separated into several MIBs. Like there is no point of walking through firewall rules part of MIB if you are having only dumb APs.

- OPENWRT-MIB - that's the place where your openwrtMIBversion and openwrtOSversion belong
- OPENWRT-WIRELESS-MIB addressing my intended scope.

Curious of your opinion.

Torstein-Eide:
Yes a MIB (the files that defined what the OID means) can be split over multiple MIB files. 
For the logic that resolves it yes it possible to multiple program, but I think a good solution would be one program, to handle all OpenWrt unique elements. 

A program will have if or case statements that gates what do to. The cost to read some more files and and format for SNMP is tiny for most low end devices, that OpenWrt runs on. LibreNMS will also reduse what is ask for from SNMP server, once a OID is add to a sensor it will only ask for that OID until next discovery. 

There is also no expatiation that you implement more that what is pressing for your need.  Ie the common + wireless stuff.

perceival:
@Torstein-Eide Your is reasonable. Going with single program and multiple MIBs aproach.

perceival:
@Torstein-Eide 

pass_persist handler is there and serves the wireless subtree live from ubus/iwinfo. Nothing per-VAP to configure. The MIB is up in librenms/librenms#19347 as OPENWRT-MIB + OPENWRT-WIRELESS-MIB. With your
original rooting — `openwrtMIB ::= { enterprises 60652 102 }` directly — also resolves a smilint descriptor clash with LIBRENMS-NOTIFICATIONS-MIB's own `librenms`.

Re the metadata would like to get your input, because one of the two is a special case:

- `distro` has to stay a net-snmp extend — as it's the OS-detection (os_detection matches nsExtendOutput."distro" containing "OpenWrt"). NMS reads it before it knows to look under .60652. An `openwrtOSVersion` scalar
  would just duplicate that, so I'd leave it out unless you want the MIB self-consistent for its own sake.
- hardware model is post-detection metadata and a good fit for a scalar (`openwrtHardwareModel`), parsed once on the device instead of per NMS.
- `openwrtMIBVersion` belongs in the MIB regardless.

Proposed: reserve `openwrtSystem = { openwrtObjects 1 }` (wireless stays at 10)
with `openwrtMIBVersion` at .1 and `openwrtHardwareModel` at .2 —
.60652.102.1.1.1.0 / .1.2.0. Does that node/OID work for you? Once it's agreed
I'll add the scalars and have the handler serve them.

Separately, a heads-up on the enterprise number: I've asked on openwrt-devel
whether OpenWrt would take an IANA PEN of its own, so the MIB can move off
LibreNMS's arc eventually. It's rooted so that's a one-line change — no dependencies.

murrant:
If we are going to use the LibreNMS PEN, I would want things to be more generic so other linux (or non-linux) os can implement them, not just OpenWRT.

perceival:
Fair point, and it's actually why we're not planning to keep this under the LibreNMS PEN. OpenWrt is registering its own enterprise number — I'm working wit OpenWrt dev on that — so none of this ends up squatting in LibreNMS's namespace. Once it's assigned the MIB re-roots onto OpenWrt's arc (a one-line change).

On genericness: you're right that the wireless table itself is vendor-neutral — it's standard 802.11 per-VAP metrics, ifIndex-joined to the interface table, nothing OpenWrt-specific in it. The only OpenWrt-flavored parts are the distro/hardware detection (which ride standard NET-SNMP-EXTEND, not the custom MIB) and the sectioned room for future firewall/dhcp arcs.

So the real question is scope: would you rather the wireless MIB be explicitly vendor-neutral so LibreNMS can point other Linux/hostapd APs at the same OIDs, even though OpenWrt authors and hosts it? I'm happy to shape it that way — the content's already generic. Just want to design the namespace once.

murrant:
Honestly, It would be great if there was a vendor neutral MIB for these things, but I fear the IEEE working group is kind of done standardizing MIBs.

I guess the question is would it be more likely to be supported by other vendors under the LibreNMS PEN or OpenWRT PEN? And would the OpenWRT devs be happy working with LibreNMS if we host it?  I'm all for whatever has the best chance of becoming a defacto standard.

I imagined LibreNMS shipping a generic Linux implementation of the MIBs that replaces this "agent" completely when I applied for the PEN.

perceival:
Update on the enterprise number: the IANA PEN request for OpenWrt is filed and open (last movement yesterday), so the re-root off LibreNMS's arc is a matter of when, not if. Nothing ends up squatting under 60652 long-term.

On which namespace becomes the de facto standard — my honest answer is neither. Near-term: OpenWrt's arc, with the table kept strictly vendor-neutral (per-VAP 802.11 metrics, ifIndex join) so any hostapd-based Linux AP can implement the same OIDs; OpenWrt is just the first implementation. Longer-term I'd like to push this as a proper management feature into a future 802.11 revision — that's the actual vendor-neutral home. I know that's a years-out play, so I'm not parking anything on it.

Practical question: are you OK merging with the provisional .60652.102 rooting as-is? The re-root is a one-line change in the MIB plus a test-data regen once the PEN lands, and I'd rather do that as a follow-up than hold both PRs on IANA's queue.

murrant:
@perceival I would prefer not to merge with the provisional PEN.  Temporary is permanant.

perceival:
Request for PEN is in the pipe. Once completed I'll ping for final review. Code review can proceed in the meantime if anyone has cycles — the re-root is a one-line MIB change, so nothing reviewed now gets invalidated.

perceival:
IANA assigned OpenWrt enterprise number **66510** today (request PHPH-3T2-M52), so the provisional rooting under the LibreNMS PEN is gone — @murrant, this is the ping I promised.

`openwrt-snmp-pass.sh` now serves the wireless subtree at `.1.3.6.1.4.1.66510.1.10`, and the README's snmpd `miboid` and snmpwalk examples match. Server side is librenms/librenms#19347, re-rooted in the same pass and green on CI.

@laf — as on the server PR, your 2026-06-20 review is still marked changes-requested but the rework you asked for landed 2026-07-05. Ready for final review on both.

uuzinger:
Tested #613's lm-sensors-pass.sh (commit 40fcbe5, fan table version) against a GL.iNet GL-BE3600 — Qualcomm IPQ53xx, pwm-fan hwmon driver, 5 tsens thermal zones. Confirmed working standalone (--snapshot) and over live pass_persist — correct Gauge32 typing for both temperature and fan tables at .1.3.6.1.4.1.2021.13.16.2.1 / .3.1.

One gap worth flagging for the LibreNMS-core side (librenms#19347): sensors() in includes/discovery/functions.inc.php only includes sensors/$class/{os_group}.inc.php or sensors/$class/{os}.inc.php — exact filename match, no fallback. Since openwrt.yaml doesn't declare os_group: unix, temperature/unix.inc.php and fanspeed/unix.inc.php never run for os: openwrt, even with this agent script correctly serving LM-SENSORS-MIB. I worked around it locally by adding temperature/openwrt.inc.php and fanspeed/openwrt.inc.php as unmodified copies of the unix versions — five temperature sensors and one fan sensor now discover correctly into Health with auto-derived thresholds.

perceival:
> One gap worth flagging for the LibreNMS-core side (librenms#19347)

Good catch on the `sensors()` mechanic — that's accurate, and `openwrt.yaml` has no `group:`, so `temperature/unix.inc.php` and `fanspeed/unix.inc.php` genuinely never fire for `os: openwrt`.

The reason you needed the workaround is that #19347 solves it a different way. It doesn't go through the `os_group` include path at all — it adds `resources/definitions/os_discovery/openwrt.yaml` with the LM-SENSORS temperature and fan tables declared as YAML discovery data. That runs via `discovery_process()`, so it's independent of the `{os_group}.inc.php` / `{os}.inc.php` lookup you traced. `tests/data/openwrt.json` in that PR has the expected output baked in (2 temp + 1 fan, `group: LM-SENSORS`, auto-derived thresholds), and the `--os o,p,q,r,s` CI job is passing.

Worth knowing: if you keep your local `temperature/openwrt.inc.php` and `fanspeed/openwrt.inc.php` and then pull #19347, you'll get each sensor twice — once as `sensor_type: lmsensors` from the includes, once as `sensor_type: openwrt` from the YAML. Drop the two copies when you update.

Thanks for the GL-BE3600 run. Five tsens zones plus a pwm-fan on IPQ53xx is a useful third platform for `lm-sensors-pass.sh`, and it's good to have the Gauge32 typing confirmed on live pass_persist rather than only `--snapshot`. Note you tested `40fcbe5`; the branch head is now `0ed376c`, which re-roots the wireless subtree onto OpenWrt's IANA PEN 66510. That doesn't touch the LM-SENSORS tables, so your results stand.

perceival:
@VVelox could you take a look at this when you get a chance? It's been open a while, CI's green, mergeable, and a couple of people (including me) have been running it in production without issues. The MIB-side counterpart is librenms/librenms#19347, also waiting on review.

perceival:
Closing this — the device-side implementation has moved to a native OpenWrt package (openwrt/packages#30547) instead of a script distributed through librenms-agent. librenms/librenms#19347 no longer depends on this PR.

PR #619, opened 2026-05-10
Adjust parsing of 'proto' to allow for "HTTP/2.0"

trurl:
In the standard Apache/nginx combined log format http2 request are logged as "HTTP/2.0" (tested with nginx), which "eq" does _not_ catch. This change checks if "proto" _contains_ "HTTP/2" or "HTTP/3".


VVelox:
thanks!

PR #627, opened 2026-08-08
Add SNMP extend script for AMD GPU power and clock

Fever-Wits:
Adds an SNMP extend script exposing AMD GPU power and clock.

These cannot reach SNMP on their own. The net-snmp `lmSensors` module maps only
the temperature, fan and voltage features of libsensors, so watts and hertz are
given no index at all and `lmMiscSensorsTable` stays empty — no configuration
changes that. GPU temperature and voltages are unaffected and keep arriving
through `LM-SENSORS-MIB` as before; this only adds what is missing.

The companion discovery side is librenms/librenms#20258.

### Output

One block of exactly four lines per card:

```
0000:c7:00.0 AMD Radeon 780M Graphics
27244000
43186000
2700000000
```

The three metric lines carry a bare number, because LibreNMS reads a sensor OID
straight into a numeric value — a label on the same line would be parsed as part
of the number and silently yield 0.

### Notes on the design

**Fixed block length.** An attribute that cannot be read is emitted as `U`, not
omitted. On APUs `power1_average` is not always available — the SMU may decline
it while idle — so a card publishing only some of the three is the ordinary
case. Omitting the line would shift every following value onto the wrong sensor;
`U` leaves a placeholder that discovery ignores. This follows `ups-nut.sh`, which
uses `Unknown` the same way.

**Cards found through the PCI driver directory**, not through `hwmonN`: hwmon
numbering is not stable across reboots, PCI slot addressing is. The glob expands
alphabetically, so block order is deterministic.

**Product names** come from the libdrm ids table, matched on device id *and*
revision — the revision matters, since e.g. `15BF` revision `C4` is a Radeon 780M
while `C5` is a 740M. A card the table does not list falls back to its PCI
address alone.

### Testing

Verified on a Ryzen 8845HS host (Radeon 780M, Proxmox), run both as root and as
the snmpd user through sudo. Also exercised against a synthetic sysfs tree with
three cards — complete, missing an attribute, and with no hwmon node — to confirm
the block length holds and a card without hwmon emits nothing rather than
shifting the others.

`shellcheck` clean at all severities; `sh -n` and `dash -n` clean.

### DIFF
--- a/snmp/amdgpu
+++ b/snmp/amdgpu
@@ -1,0 +1,319 @@
+#!/bin/sh
+################################################################
+# Instructions:                                                #
+# 1. copy this script to /etc/snmp/ and make it executable:    #
+#    chmod +x amdgpu                                           #
+# 2. edit your snmpd.conf to include this line:                #
+#    pass_persist .1.3.6.1.4.1.60652.101 /etc/snmp/amdgpu      #
+# 3. restart snmpd on the host                                 #
+################################################################
+#
+# Serves LIBRENMS-AMDGPU-MIB: the power and clock of the AMD GPUs in a host,
+# which cannot reach SNMP on their own. The net-snmp lmSensors module maps only
+# the temperature, fan and voltage features of libsensors, so watts and hertz
+# are given no index at all. Temperature and voltages keep arriving through
+# LM-SENSORS-MIB as before; this script adds only what is missing.
+#
+# One row per card, one column per metric:
+#
+#   .1.3.6.1.4.1.60652.101.1.1.<column>.<index>
+#
+#   column 2  amdGpuProductName   string  name from the libdrm ids table
+#   column 3  amdGpuPower         gauge   power1_input, microwatts
+#   column 4  amdGpuPowerAverage  gauge   power1_average, microwatts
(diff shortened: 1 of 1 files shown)


Fever-Wits:
Reworked the output format after finding that the consumer side had to rebuild the card blocks by counting lines.

Each card now emits an identifier line naming the metrics it carries, followed by those metrics in that order:

```
0000:c7:00.0 power1_average,power1_input,freq1_input AMD Radeon 780M Graphics
30215000
40067000
2700000000
```

The names live on the identifier line because the value lines cannot hold them — LibreNMS reads a sensor OID straight into a numeric value, so a label beside the number is parsed as part of it.

The metric list is built from the same variable the loop reads, so a metric added to the script appears in both places or in neither.

The consumer side is librenms/librenms#20258; the two go together.

Fever-Wits:
Rewritten as `pass_persist` — it now serves a table under `.1.3.6.1.4.1.60652.101` instead of printing lines, so the consumer side no longer has to rebuild the blocks. Indexed by PCI address; a reading the card does not publish has no cell at all rather than a placeholder.

Details and the reasoning are on the LibreNMS side: librenms/librenms#20258. The two go together — this OID space is not usable without the MIB that ships there.

PR #629, opened 2026-08-18
Add AdGuard Home snmp extend

gthelding:
SNMP extend for [AdGuard Home](https://github.com/AdguardTeam/AdGuardHome).

Python 3, stdlib only. Reads `/etc/snmp/adguard.json` (API URL + web UI credentials), queries `/control/status` and `/control/stats`, and prints LibreNMS application JSON v1.

Extend name must be `adguard`. Stats are AdGuard's rolling window (24h by default), not monotonic counters. Default HTTP timeout is 5s per call (two calls per poll); the SNMP timeout must exceed `2 * timeout`.

Companion LibreNMS application PR: https://github.com/librenms/librenms/pull/20339

### DIFF
--- a/snmp/adguard
+++ b/snmp/adguard
@@ -2,7 +2,8 @@
 """LibreNMS snmp extend script for AdGuard Home.
 
 Queries the AdGuard Home REST API (/control/status and /control/stats) and
-prints the results as LibreNMS application JSON (version 1).
+prints the results as LibreNMS application JSON (version 1),
+gzip+base64 encoded so snmpd does not mangle the payload.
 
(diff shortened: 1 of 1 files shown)


snmp/adguard
VVelox, line 1: You likely want to add in gzip+base64 encapsulation.

Raw JSON can play havor with some net-snmp snmpd versions.

VVelox (changes requested):

VVelox:
Looks good. If you could get gzip+base64 added, that would be great. :)

gthelding:
done, gzip+base64 in output()

VVelox:
Awesome, thanks!

PR #630, opened 2026-08-18
Default AdGuard HTTP timeout to 1 second

gthelding:
Follow-up to #629.

LibreNMS defaults to a 1s SNMP timeout (`snmp.timeout`). The AdGuard extend makes two HTTP calls, so a 5s per-call default will generally miss the poll if AdGuard hangs.

Drop the default to 1s. Local AdGuard answers in milliseconds. If the API is remote or slow, cache via cron and `cat` the file (documented on librenms/librenms#20339) instead of raising the SNMP timeout.


VVelox:
thanks!

PR #631, opened 2026-08-19
add crs125/rb1100 vlan support

si458:


si458:
changed to draft as it worked on my crs125 but then broke my rb1100 because the rb1100 doesnt support switching vlans

bennet-agital:
```
:do { :put [/interface ethernet switch vlan print count-only] } on-error={ :put "unsupported" }
```

if that works and outputs "unsupported" you can use the following conditioning:

```
:local hasswvlan true
:do {
    :put [/interface ethernet switch vlan print count-only]
} on-error={ :set hasswvlan false }

:if ($hasswvlan) do={
    :foreach i in [/interface ethernet switch vlan find] do={
        ... your patch
    }
}
```

can you test this on your rb1100? I don't have a device without vlan support.

si458:
thanks @bennet-agital that indeed works!
now both my rb1100 and crs125 show vlans!

PR #633, opened 2026-09-03
feat: add multi-distro reboot-required SNMP extend script

ahoogerhuis:
The extend script consumed by librenms/librenms#20446 (the companion application-module PR on the core repo). Self-detects distro family via `/etc/os-release` (`$ID`/`$ID_LIKE`) and delegates to the appropriate native check per family, rather than shipping separate scripts.

Debian/Ubuntu via `/var/run/reboot-required`, RHEL/CentOS/Rocky/Alma/Fedora/Amazon Linux via `needs-restarting -r`, SUSE/SLES/openSUSE via `zypper needs-rebooting`, Arch via comparing the running kernel against installed modules. Exits non-zero with no output on an unrecognized distro or missing tool, rather than guessing.

On AI assistance: this script involved AI-assisted development (Claude, plus Claude Code), reviewed and verified by me before submission.


bennetgallein (approved):

bennetgallein (changes requested):
should switch to emitting json. Probably not "easy" in bash (doable) but a short python version is probably prefered

ahoogerhuis:
> should switch to emitting json. Probably not "easy" in bash (doable) but a short python version is probably prefered

One rewrite coming up shortly.

PR #634, opened 2026-09-06
fix(zfs): dynamic header parsing, ns/ms conversion, live-sample interval

ahoogerhuis:
Three related bugs in the `zfs` extend script's `zpool` parsing, found while adding a new check:

`zpool iostat` with no interval returns a since-boot average, not a live reading -- scrub/trim wait graphs never dropped to zero after activity ended. Fixed with a real one-shot interval sample.

The fixed-position field parser assumed a specific column count. Current OpenZFS adds columns (`rebuild_wait` and others), and Perl's positional list assignment meant everything after the insertion point silently read its neighbor's value. Replaced with parsing based on the real column headers, so future OpenZFS versions that add/reorder columns get handled correctly instead of silently misaligning.

Time values are documented as nanoseconds; nothing converted them, so every wait-time graph was off by ~1,000,000x. Fixed.

Also renamed the script from `zfs` to `zfs-stats` -- the old name would land at a path like `/etc/snmp/zfs` on installation, shadowing the real `zfs` binary on `$PATH`.

Verified against real captured `zpool` output (both known column-count layouts) and a synthetic layout with a new, previously-unseen column, confirming the parser resolves fields by name correctly in a case the old positional parser would have silently gotten wrong. Also deployed and confirmed on a real production host.

Developed with Claude's assistance; I've reviewed and take responsibility for the change.


ahoogerhuis:
There is some more bugs in the L2 section of this script, so this is not worth merging just yet.

ahoogerhuis:
A few squashed, but there is one unexplained nan left, and need to engineer a way to wiggle all the error/rebuild related values to see if they work.

snmp/zfs-stats
```diff
@@ -221,6 +352,18 @@ while ( defined( $pools[$pools_int] ) ) {
 		}
 	}
 
+	# zpool iostat -p reports time values in nanoseconds (per zpool-iostat(8)),
```
VVelox, line 355: This should be removed. Graph unit should be fixed.

snmp/zfs-stats
```diff
@@ -191,24 +290,56 @@ while ( defined( $pools[$pools_int] ) ) {
...
-		$parsed,                  $parsed,                 $parsed,                 $newPool{operations_r},
-		$newPool{operations_w},   $newPool{bandwidth_r},   $newPool{bandwidth_w},   $newPool{total_wait_r},
-		$newPool{total_wait_w},   $newPool{disk_wait_r},   $newPool{disk_wait_w},   $newPool{syncq_wait_r},
-		$newPool{syncq_wait_w},   $newPool{asyncq_wait_r}, $newPool{asyncq_wait_w}, $newPool{scrub_wait},
-		$newPool{trim_wait},      $newPool{syncq_read_p},  $newPool{syncq_read_a},  $newPool{syncq_write_p},
-		$newPool{syncq_write_a},  $newPool{asyncq_read_p}, $newPool{asyncq_read_a}, $newPool{asyncq_write_p},
-		$newPool{asyncq_write_a}, $newPool{scrubq_read_p}, $newPool{scrubq_read_a}, $newPool{trimq_write_p},
-		$newPool{trimq_write_a},
-	) = split( /\,/, $iostat );
+	# -H is dropped here too, for the same reason as the pool-list
```
VVelox, line 293: Dumbass garbage comment added by a shitty LLM.

snmp/zfs-stats
```diff
@@ -191,24 +290,56 @@ while ( defined( $pools[$pools_int] ) ) {
...
+
+	my %iostat_raw;
+	foreach my $i ( 0 .. $#$iostat_field_names ) {
+		my ( $s, $e ) = @{ $iostat_field_ranges->[$i] };
+		my $v = substr( $iostat_data_line, $s, $e - $s );
+		$v =~ s/^\s+|\s+$//g;
+		$iostat_raw{ $iostat_field_names->[$i] } = $v;
+	}
+
+	# The script's own internal field names, aliased from the real
```
VVelox, line 317: Dumbass garbage comment added by a shitty LLM.

snmp/zfs-stats
```diff
@@ -472,6 +615,22 @@ foreach my $item (@l2_keys) {
 $tojson{l2_errors}       = $tojson{l2_writes_error} + $tojson{l2_cksum_bad} + $tojson{l2_io_error};
 $tojson{l2_access_total} = $tojson{l2_hits} + $tojson{l2_misses};
 
+# LibreNMS core (includes/polling/applications/zfs.inc.php) looks up these
```
VVelox, line 618: Rambling.

snmp/zfs-stats
```diff
@@ -472,6 +615,22 @@ foreach my $item (@l2_keys) {
...
+# two fields under abbreviated names that don't match the real kernel ARC
+# stat names used above -- alias them alongside the full names rather than
+# renaming, so this script's own output stays self-describing while also
+# satisfying core's mismatched expectation.
+$tojson{l2_bufc_d_asize} = $tojson{l2_bufc_data_asize}     if defined $tojson{l2_bufc_data_asize};
+$tojson{l2_bufc_m_asize} = $tojson{l2_bufc_metadata_asize} if defined $tojson{l2_bufc_metadata_asize};
+
+# Core also expects this pre-computed ratio, which nothing above ever
```
VVelox, line 626: Dumbass garbage comment added by a LLM.

snmp/zfs-stats
```diff
@@ -122,8 +214,19 @@ if ($help) {
 }
 
 #process each pool and shove them into JSON
-my $zpool_output = `/sbin/zpool list -pH`;
-my @pools        = split( /\n/, $zpool_output );
+#
```
VVelox, line 217: More dumbass garbage comments added by a LLM.

snmp/zfs-stats
```diff
@@ -2,16 +2,31 @@
...
 =head1 VERSION
 
-0.3.0
+0.4.0
 
 =head1 DESCRIPTION
 
 For more information, see L<https://docs.librenms.org/Extensions/Applications/#zfs>.
 
+=head1 WHY "zfs-stats" AND NOT "zfs"
```
VVelox, line 15: Holy goatse'... no...

Lets not re-name this.

Also definitely does not belong under bin.

snmp/zfs-stats
```diff
@@ -29,7 +44,12 @@ in the return.
 
 =head1 SNMPD SETUP EXAMPLES
 
-    extend zfs /etc/snmp/zfs
+The extend name (C<zfs> below) is just a label LibreNMS uses
```
VVelox, line 47: This is wrong and rambling.

VVelox (changes requested):
Don't rename this and the reason given is bad.

The commends added are nearly pure garbage.

Switching to using headers is not bad bad idea.

The NS thing would be best fixed in the graphs and not here.

VVelox:
If using a LLM, please pay a hell of a lot more attention to it given the shear amount of garbage they will output.

ahoogerhuis:
@VVelox Love your comment styling.

PR #635, opened 2026-09-06
Fix SMART temperature parsing in smart-v1

githonk:
Fixes two issues in `snmp/smart-v1` related to SMART temperature handling:

1. Fix `drive_drip` typo: should be `drive_trip`.

2. Fix `max_temp` selection logic in the three cases selecting the maximum temperature from SMART attributes 190 and 194.

The `defined()` calls incorrectly contain `&&` expressions. The `&&` expression is evaluated first and its result is passed to `defined()`, rather than testing whether the individual value is defined and numeric.

The conditions are corrected so that the presence and numeric validity of SMART attributes 190 and 194 are tested independently.

No other SMART parsing logic is changed.


VVelox:
derp, thanks

PR #636, opened 2026-09-11
add vminfo discovery script

bennetgallein:
adds a new cache-based vminfo collection script to include Proxmox VMs & LXC containers into the /vminfo page

I'm no python expert so this has been helped by Mr Claude, open to feedback


bennetgallein:
I dropped ostype detection because it does not make much sense, you select a kernel range or windows generation... This might allow us to drop the whole caching logic, but i'll keep it in for now. On larger clusters this might come in handy

